PR #20, opened 2022-09-07
Add first example

tcompa:
By now this PR is for a trivial example on fake data, but we can later include a second example using the tiny public dataset (to be added, see https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/29).


tcompa:
My bad, I was adding examples to the wrong repo. Closing this PR, in favor of a new one on the client repo.

Issue #4, opened 2022-09-01
Define/implement standard parsl executors for fractal


tcompa:
Merging https://github.com/fractal-analytics-platform/fractal/pull/172 will close fractal-analytics-platform/fractal-server#8, so that we now can specify a given executor for a task (in the manifest, or by modifying the task later on). The list of allowed executors is currently a trivial one, only including `cpu` and `cpu-2` (which are actually identical).

We should discuss on:
1. Which list of standard executors would we include? There could be four combination of `{high,low}-{mem,cpu}-cpu`, and at least one `gpu` one. There can be also a couple of gpu executors, like `gpu-single` (limited to a single gpu node) and `gpu-many`.. or something like this.
2. Where will this information on standard executors be stored (as an hard-coded list of parameters)? There are simple answers to this question, but we should first understand how many of these standards should exist: a unique one? one per lab?
3. How could a user define a fully-custom executor? This is currently not implemented. We should probably move it to another issue, after an assessment of how important it is.

tcompa:
I am not satisfied with the current approach in https://github.com/fractal-analytics-platform/fractal-server/pull/24.

Defining a set of executors (with their given labels, like `cpu`, `gpu`, ...) each time a workflow is submitted looks dangerous. If a user submits two workflows in a row, upon submission of the second one there will be a `parsl.clear()` statement, and a new `parsl.config.Config(...)` one. This means that whatever was provided as a first config is removed and replaced by the new one.

I'm leaning towards a new, more global, approach. The server has a single set of executors, which is defined on startup. The specifics of these executors depend on an environment variable `FRACTAL_PARSL_CONFIG`, which can be set for instance to `local`, or `pelkmanslab` or `custom`. Loading a config from a set of pre-defined ones is the standard behavior (e.g. `local` will point towards two local executors, `cpu` and `cpu-2`, with the second one being useful for testing), while defining a custom one is not currently supported and will be defined later (relevant point: if executors are defined and initialized at server startup, the user cannot really modify with that.. this has to be re-assessed later, possibly leading to not letting the user define their own executors).

This new scheme is somewhat logical: a SLURM cluster comes with a set of (pre-defined) executors available, and users submit their tasks to the relevant ones. 
I will test it and see how it works.

tcompa:
Another solution would be to internally replace the executor names (e.g. `cpu`) with something workflow-related (e.g. `cpu | my worfklow name`).

jluethi:
[...]

PR #24, opened 2022-09-08
Parsl executors

tcompa:
This PR addresses #4

### DIFF
--- a/fractal_server/app/runner/__init__.py
+++ b/fractal_server/app/runner/__init__.py
@@ -24,6 +24,7 @@ from ..models.task import Task
 from .runner_utils import add_prefix
 from .runner_utils import async_wrap
 from .runner_utils import load_parsl_config
+from .runner_utils import shutdown_executors
 
 
 def _task_fun(
@@ -378,6 +379,8 @@ async def submit_workflow(
     output_dataset.meta = await async_wrap(get_app_future_result)(
         app_future=final_metadata
     )
+
+    shutdown_executors(workflow_id=workflow.id)
 
     db.add(output_dataset)
 
--- a/fractal_server/app/runner/__init__.py
+++ b/fractal_server/app/runner/__init__.py
@@ -9,13 +9,9 @@ from typing import Literal
 from typing import Optional
 from typing import Union
 
-import parsl
-from parsl.addresses import address_by_hostname
 from parsl.app.app import join_app
 from parsl.app.python import PythonApp
-from parsl.config import Config
 from parsl.dataflow.futures import AppFuture
-from parsl.monitoring.monitoring import MonitoringHub
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from ...config import settings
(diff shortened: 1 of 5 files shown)


tcompa:
Quoting from https://github.com/fractal-analytics-platform/fractal-server/issues/4#issuecomment-1241794819

> I am not satisfied with the current approach in #24.
> 
> Defining a set of executors (with their given labels, like `cpu`, `gpu`, ...) each time a workflow is submitted looks dangerous. If a user submits two workflows in a row, upon submission of the second one there will be a `parsl.clear()` statement, and a new `parsl.config.Config(...)` one. This means that whatever was provided as a first config is removed and replaced by the new one.
> 
> I'm leaning towards a new, more global, approach. The server has a single set of executors, which is defined on startup. The specifics of these executors depend on an environment variable `FRACTAL_PARSL_CONFIG`, which can be set for instance to `local`, or `pelkmanslab` or `custom`. Loading a config from a set of pre-defined ones is the standard behavior (e.g. `local` will point towards two local executors, `cpu` and `cpu-2`, with the second one being useful for testing), while defining a custom one is not currently supported and will be defined later (relevant point: if executors are defined and initialized at server startup, the user cannot really modify with that.. this has to be re-assessed later, possibly leading to not letting the user define their own executors).
> 
> This new scheme is somewhat logical: a SLURM cluster comes with a set of (pre-defined) executors available, and users submit their tasks to the relevant ones. I will test it and see how it works.

tcompa:
As of https://github.com/fractal-analytics-platform/fractal-server/pull/24/commits/ec3df97d94cae4d3c66dc2ed3d4c66611174185b, we have non-trivial examples that run reasonably well (*) both locally and on pelkmanslab (see https://github.com/fractal-analytics-platform/fractal-server/issues/4#issuecomment-1243407277). I propose that we review this PR and merge to `dev`/`main` (also leading to a MINOR or PATCH release).

Known issues about this PR:
1. Parsl information (monitoring and logs) is less transparent, as it is not any more workflow-specific;
2. The number of executors is not limited;
3. We are not yet handling the (edge) case where two different workflows have the same ID.
4. This PR does not define the specifics of the `{low,high}-{mem,cpu}` executors on pelkmans, but only prepares the architecture for doing so.

On the other hand, this PR finally allows to easily use a set of pre-defined parsl executors (e.g. `local` or `pelkmanslab`, and in principle also a `custom` one although it is not yet implemented).
Note that earlier version were working only because in the examples we would typically submit workflows one by one. When I tried submitting several of them at the same time, they got assigned to the wrong executor (see https://github.com/fractal-analytics-platform/fractal-server/issues/4#issuecomment-1241794819).

fractal_server/app/runner/__init__.py
```diff
@@ -240,9 +171,16 @@ def _atomic_task_factory(
...
         task_executor = settings.PARSL_DEFAULT_EXECUTOR
-    if task_executor not in valid_executors:
+    task_executor = add_prefix(
+        workflow_id=workflow_id, executor_label=task_executor
+    )
+
+    # Verify match between task_executor and available executors
+    valid_executor_labels = DataFlowKernelLoader.dfk().executors.keys()
+    if task_executor not in valid_executor_labels:
         raise ValueError(
```
jacopo-exact, line 181: Non blocking: I think we should spin off everything that concerns setting up executors and parsl related stuff into its own routines and keep only a single call in `_atomic_task_factory`. As the name suggests, that function should only be concerned with preparing the tasks themselves, irrespectively of where they run.

in any case we can take care of separation of concerns in a subsequent refactoring.

fractal_server/config.py
```diff
@@ -114,11 +114,10 @@ def DB_ECHO(self):
...
     ###########################################################################
     DATA_DIR_ROOT: str = fail_getenv("DATA_DIR_ROOT")
-    USE_SLURM: bool = bool(int(getenv("USE_SLURM", "0")))
+    PARSL_CONFIG: str = getenv("PARSL_CONFIG", "local")
     PARSL_DEFAULT_EXECUTOR: str = getenv("PARSL_DEFAULT_EXECUTOR", "cpu")
-    if USE_SLURM:
-        SLURM_PARTITION_CPU: str = fail_getenv("SLURM_PARTITION_CPU")
-        # FIXME add gpu partition
+    if PARSL_CONFIG == "custom":
+        PARSL_CONFIG_CUSTOM_FILE: str = fail_getenv("PARSL_CONFIG_CUSTOM_FILE")
```
jacopo-exact, line 120: what is this supposed to be? in the prospect of shipping fractal into a container, I would advise against configuration files and keep everything in the environment, which makes it extremely easier to handle in docker / k8s and the like.

in any case It seems like this variable is defined but never used. If that's the case, please remove it until we actually need it

jacopo-exact (commented):
overall good to go. please review comments, although no real need to address them at this time.

jacopo-exact (approved):

fractal_server/app/runner/__init__.py
tcompa: Fixed (?) with 09a5e798ab1a9de24a15c984119e254f08a24fa5. Now it looks like:
https://github.com/fractal-analytics-platform/fractal-server/blob/09a5e798ab1a9de24a15c984119e254f08a24fa5/fractal_server/app/runner/__init__.py#L149-L173

tcompa:
Thanks Jacopo for your comments, I'm merging.

tcompa:
Detailed configuration will be defined in https://github.com/fractal-analytics-platform/fractal-server/issues/34.

I'm closing this, after repeating the main open questions (which can be used for new issues):
* Parsl information (monitoring and logs) is less transparent, as it is not any more workflow-specific;
* The number of executors is not limited;
* We are not yet handling the (edge) case where two different workflows have the same ID.

PR #29, opened 2022-09-12
Include python 3.9 for github CI

### LINKED ISSUE #7: Python version (3.8 or 3.9)
Does fractal install/work correctly also on python 3.9 (on top of 3.8, which is known to work)?

If so, should we also add the corresponding tests to the Github CI?

tcompa:
closes #7


tcompa:
Github CI is passing for both python 3.8 and 3.9, I'm merging this into `dev`.

PR #80, opened 2022-09-26
10 how does fractal submit slurm jobs from the correct user

jacopo-exact:


jacopo-exact:
@jluethi @tcompa @mfranzon
The promised update on how this PR is proceeding. Unfortunately we encountered a new road block which is quite difficult to bypass, having to do with the intricacies of how jobs are handled by Parsl.

Will keep going at it for the remaining of today and tomorrow.

jluethi:
Ok, thanks for the update @jacopo-exact

One high level question: With the new architecture, can we still run the server on the cluster head node and submit to it via a bash script that defines a user with username & password? Because that is a) very convenient for testing and b) may also be how we'd run demos to keep it simple for the time being.

Also, another high-level feedback from @gusqgm : FMI IT seems less concerned with having a sudo user that submits on the cluster head node than having ssh keys. I'm sure we'll find a way to get a setup with ssh keys if that is the best way. But if it gets much harder to do things with ssh tunneling, let's reconsider just using a sudo user.

jacopo-exact:
> One high level question: With the new architecture, can we still run the server on the cluster head node and submit to it via a bash script that defines a user with username & password? Because that is a) very convenient for testing and b) may also be how we'd run demos to keep it simple for the time being.

username and passwords is the current way the user authenticates with the fractal-server, and this won't change for the time being. The ssh transport is the way that fractal-server communicates with the slurm daemon. 

> Also, another high-level feedback from @gusqgm : FMI IT seems less concerned with having a sudo user that submits on the cluster head node than having ssh keys. I'm sure we'll find a way to get a setup with ssh keys if that is the best way. But if it gets much harder to do things with ssh tunneling, let's reconsider just using a sudo user.

The sudo way has its long list of technical difficulties that are solved by the ssh transport. There should be no ssh tunneling involved in the communication between fractal-client and fractal-server. If this is the issue you're facing, we'll set up a machine exclusively dedicated to serving the fractal-server on a standard port (80 or 443).

jluethi:
Sounds reasonable to me, thanks for the context @jacopo-exact !
Also, if we use username & password or some oauth2 credentials later, there shouldn't be ssh tunneling between server & client, I agree. So no need to set up an additional machine for that for the moment :)

tcompa:
Closed in favor of #112

PR #88, opened 2022-09-30
Dfk runners

### LINKED ISSUE #83: garcefully shutdown parsl executors at the end of workflow
@tcompa I turned the feature back on in PR #80 as otherwise the pipeline fails, likely because of too many executors hanging and clogging all available sockets.

jacopo-exact:
A dedicated Parsl DataFlowKernel (DFK) is created for each workflow.
The task factory is refactored to set explicitly the DFK the app it creates belongs to.

close #83, #3


tcompa:
This PR is almost ready, and quite valuable in terms of server stability (and non-proliferation of processes), on top of giving a much better organized parsl monitoring.
A major problem is still open though:
- [ ] After a workflow completes, the cleanup of parsl DataFlowKernel also shuts down the server.

Issue #141, opened 2022-10-13
worfklow api


jacopo-exact:
implement tests + full crud for the Workflow object.

- [x] `POST /workflow/`
- [x] `DELETE /workflow/{_id}`
- [ ] get `GET /workflow/{_id}`
- [ ] get list `GET /project/{_id}/workflows/`
- [ ] `PATCH /workflow/{_id}/add-task/`
- [ ] remove task `DELETE /workflow/{_id}/rm-task/{task_id}`
- [ ] edit `PATCH /workflow/{_id}` (except for task list attribute)

please open PR against main branch

PR #155, opened 2022-10-19
workflow api POST

ychiucco:
ref #141

### DIFF
--- a/fractal_server/app/api/v1/workflow.py
+++ b/fractal_server/app/api/v1/workflow.py
@@ -14,34 +14,45 @@ from fastapi import APIRouter
 from fastapi import Depends
 from fastapi import HTTPException
 from fastapi import status
-from sqlalchemy.exc import NoResultFound
 from sqlmodel import select
 
 from ...db import AsyncSession
 from ...db import get_db
-from ...security import current_active_user
-from ...security import User
-
 from ...models import Workflow
 from ...models import WorkflowCreate
 from ...models import WorkflowRead
+from ...security import current_active_user
+from ...security import User
+from .project import get_project_check_owner
 
 router = APIRouter()
 
 
-@router.post("/", response_model=WorkflowRead, status_code=status.HTTP_201_CREATED)
+@router.post(
+    "/", response_model=WorkflowRead, status_code=status.HTTP_201_CREATED
+)
 async def create_workflow(
     workflow: WorkflowCreate,
     user: User = Depends(current_active_user),
     db: AsyncSession = Depends(get_db),
 ):
+    await get_project_check_owner(
+        project_id=workflow.project_id,
+        user_id=user.id,
+        db=db,
+    )
     # Check that there is no workflow with the same name
-    stm = select(Workflow).where(Workflow.name == workflow.name)
+    stm = (
+        select(Workflow)
+        .where(Workflow.name == workflow.name)
+        .where(Workflow.project_id == workflow.project_id)
+    )
     res = await db.execute(stm)
     if res.scalars().all():
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            detail=f"Workflow name ({workflow.name}) already in use",
+            detail=f"Workflow with name={workflow.name} and\
+                    project_id={workflow.project_id} already in use",
         )
     db_workflow = Workflow.from_orm(workflow)
     db.add(db_workflow)
--- a/fractal_server/app/api/v1/workflow.py
+++ b/fractal_server/app/api/v1/workflow.py
@@ -42,6 +42,7 @@ async def create_workflow(
         db=db,
     )
     # Check that there is no workflow with the same name
+    # and same project_id
     stm = (
         select(Workflow)
         .where(Workflow.name == workflow.name)
--- a/fractal_server/app/api/v1/workflow.py
+++ b/fractal_server/app/api/v1/workflow.py
@@ -12,6 +12,10 @@ Zurich.
 """
 from fastapi import APIRouter
 from fastapi import Depends
+from fastapi import HTTPException
+from fastapi import status
+from sqlalchemy.exc import NoResultFound
+from sqlmodel import select
 
 from ...db import AsyncSession
 from ...db import get_db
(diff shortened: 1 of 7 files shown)


tests/test_workflow_api.py
```diff
@@ -0,0 +1,31 @@
...
+from fractal_server.app.models import Task
+from fractal_server.app.models import TaskCreate
+from fractal_server.tasks import collect_tasks
+
+async def test_workflow_post(db, client, MockCurrentUser):
+    fake_workflow = {
+        "name": "My Workflow",
+        "resource_type": "core task",
+        "input_type": "Boh",
+        "output_type": "Boh",
```
jacopo-exact, line 14: these parameters are not part of the Workflow schema, cf. `fractal_server/app/schames/task.py` and `fractal_server/app/models/workflow.py`

in particular, we should test that if a workflow is defined without providing a project id, the request fails.

tests/test_workflow_api.py
```diff
@@ -0,0 +1,31 @@
...
+        )
+        res2 = await client.post(
+            "api/v1/workflow/",
+            json=fake_workflow,
+        )
+        assert res1.status_code==201
+        assert res2.status_code==422
+    
+    debug(res1, res1.json())
+    debug(res2, res2.json())
```
jacopo-exact, line 30: as a common guideline, write each assertion immediately after the call it tests:
```
res1 = ...
assert res1

res2 = ...
assert res2
```

tests/test_workflow_api.py
```diff
@@ -0,0 +1,31 @@
...
+        res1 = await client.post(
+            "api/v1/workflow/",
+            json=fake_workflow,
+        )
+        res2 = await client.post(
+            "api/v1/workflow/",
+            json=fake_workflow,
+        )
+        assert res1.status_code==201
+        assert res2.status_code==422
```
jacopo-exact, line 27: these assertion are weak. you should test that the object is actually persisted to the db and that the relationship with a project is actually established.

fractal_server/app/api/v1/workflow.py
```diff
@@ -52,3 +71,6 @@ async def get_list_workflow(
     List public workflows
     """
     raise NotImplementedError
+
+
+### ----------
```
jacopo-exact, line 76: keep code clean: do not insert placeholder comments, as they rapidly make their way in the stable codebase, decreasing readability and overall code quality

fractal_server/app/api/v1/workflow.py
```diff
@@ -12,34 +12,53 @@
...
 from ...db import get_db
 from ...security import current_active_user
 from ...security import User
 
+from ...models.workflow import Workflow
+
 router = APIRouter()
 
 
 @router.post("/")
```
jacopo-exact, line 30: missing return model. it should be `WorkflowRead`

fractal_server/app/api/v1/workflow.py
```diff
@@ -12,34 +12,53 @@
...
 async def create_workflow(
+    workflow: Workflow,
```
jacopo-exact, line 32: wrong input model: it should be `WorkflowCreate`. If it doesn't exist yet, please define it.

fractal_server/app/api/v1/workflow.py
```diff
@@ -12,34 +12,53 @@
...
     """
-    raise NotImplementedError
+    # Check that there is no workflow with the same name
+    stm = select(Workflow).where(Workflow.name == workflow.name)
+    res = await db.execute(stm)
+    if res.scalars().all():
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail=f"Workflow name ({workflow.name}) already in use",
+        )
```
jacopo-exact, line 46: this was not a requirement but you have a good point. When is it that we cannot reuse the same workflow name? I think it should be on a per project basis (please @tcompa  comment on this).

pending Tom's comment, please only check that the same name is not taken within the same project rather than globally.

Note: if we needed a unique constraint globally, we would implement it at the mapper level.

fractal_server/app/api/v1/workflow.py
```diff
@@ -12,34 +12,53 @@
...
+    db_workflow = Workflow.from_orm(workflow)
+    db.add(db_workflow)
+    await db.commit()
+    await db.refresh(db_workflow)
+    return db_workflow
 
 
-@router.post("/{wofklow_id}")
+@router.post("/{workflow_id}")
```
jacopo-exact, line 54: this does not conform to spec. As per #141, `PATCH /workflow/{_id}/add-task/`

jacopo-exact (changes requested):
please address comments

fractal_server/app/api/v1/workflow.py
tcompa: There can exist two workflows with the same name, but only if they are associated to different projects

tests/test_workflow_api.py
ychiucco: If I do
```
res = await client.post(
            "api/v1/workflow/",
            json={
                "name": "My Workflow",
                "project_id": 1,
            },
)
debug(res.json())
```
I get
```
res.json(): {
        'detail': [
            {
                'loc': [
                    'body',
                    'resource_type',
                ],
                'msg': 'field required',
                'type': 'value_error.missing',
            },
            {
                'loc': [
                    'body',
                    'input_type',
                ],
                'msg': 'field required',
                'type': 'value_error.missing',
            },
            {
                'loc': [
                    'body',
                    'output_type',
                ],
                'msg': 'field required',
                'type': 'value_error.missing',
            },
        ],
    }
```
Why does it require these fields if they are not part of Workflow?

tests/test_workflow_api.py
```diff
@@ -0,0 +1,74 @@
...
+
+        stm = select(LinkTaskWorkflow)
+        _link = await db.execute(stm)
+        link_list = _link.scalars().all()
+        
+        for link in link_list:
+            assert link.workflow_id==w_id
+            assert link.task_id in task_id_list
+            task_id_list.remove(link.task_id)
+        assert task_id_list == []
```
jacopo-exact, line 74: this test concerns the creation of a `Workflow`. Here you are creating tasks and adding them to the workflow, but that is not what you want to test.

Instead you should check that the db contains a workflow with the correct data you passed to the endpoint (name and project_id)

tests/test_workflow_api.py
```diff
@@ -0,0 +1,74 @@
...
+            json={"name":"My Workflow"}, 
+        )
+        assert res.status_code==422 # no project_id
+
+        project = await project_factory(user)
+        p_id = project.id
+        workflow = WorkflowCreate(
+            name="My Workflow",
+            project_id=p_id,
+        )
```
jacopo-exact, line 29: This is correct. However, it is preferable that you use plain dictionaries to pass data to the endpoints. For example, if in the future we change the `WorkflowCreate` schema, the class might silently add defaults and the test still silently pass even though we need to explicitly pass some extra parameter to it.

a simple change `WorkflowCreate -> dict` will make us a bit safer in the future.

jacopo-exact (changes requested):
Still a few comments to address but we are almost there.

jacopo-exact (approved):
perfect!

PR #160, opened 2022-10-20
DELETE workflow

ychiucco:
ref #141


tests/test_workflow_api.py
```diff
@@ -60,3 +59,44 @@ async def test_workflow_post(
...
+        workflow = {
+            "name": "My Workflow",
+            "project_id": p_id,
+        }
+
+        res = await client.post(
+            "api/v1/workflow/",
+            json=workflow,
+        )
+        wf = (await db.execute(select(Workflow))).scalars().one()
```
jacopo-exact, line 85: you already know the `workflow.id`, so you can use directly await db.get() and simplify the logic

tests/test_workflow_api.py
```diff
@@ -60,3 +59,44 @@ async def test_workflow_post(
...
+        t0 = await task_factory()
+        t1 = await task_factory()
+        await wf.insert_task(t0, db=db)
+        await wf.insert_task(t1, db=db)
+
+        assert len((await db.execute(select(Workflow))).scalars().all()) == 1
+        assert (
+            len((await db.execute(select(WorkflowTask))).scalars().all()) == 2
+        )
```
jacopo-exact, line 94: it would be better to query only the `WorkflowTask`s belonging to the `Workflow` you are dealing with. For instance, this would allow to also test that other `WorkflowTask`s are not touched by the delete.

tests/test_workflow_api.py
```diff
@@ -60,3 +59,44 @@ async def test_workflow_post(
...
+
+        res = await client.delete(f"api/v1/workflow/{wf.id}")
+        assert res.status_code == 204
+
+        assert len((await db.execute(select(Workflow))).scalars().all()) == 0
+        assert (
+            len((await db.execute(select(WorkflowTask))).scalars().all()) == 0
+        )
```
jacopo-exact, line 102: as per previous comment, it would be better to use specific queries

jacopo-exact (changes requested):

jacopo-exact (approved):

fractal_server/app/api/v1/workflow.py
```diff
@@ -60,3 +61,20 @@ async def create_workflow(
...
     await db.refresh(db_workflow)
     return db_workflow
+
+
+@router.delete("/{_id}", status_code=status.HTTP_204_NO_CONTENT)
+async def delete_workflow(
+    _id: int,
+    user: User = Depends(current_active_user),
+    db: AsyncSession = Depends(get_db),
+):
```
jacopo-exact, line 71: missing access control.
please check that user has rights on the project the workflow belongs to. raise the appropriate error otherwise

fractal_server/app/api/v1/workflow.py
```diff
@@ -60,3 +61,20 @@ async def create_workflow(
...
+    workflow = await db.get(Workflow, _id)
+    if not workflow:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail="Workflow not found"
+        )
+    await db.delete(workflow)
+    await db.commit()
+    workflow = await db.get(Workflow, _id)
```
jacopo-exact, line 79: ??
unused variable. why didn't the linter catch this error?

jacopo-exact (changes requested):

jacopo-exact (approved):

PR #164, opened 2022-10-20
unit tests for cascade on delete

### LINKED ISSUE #156: cascade for Project-Workflow and Workflow-WorkflowTask
- [ ] when a project is deleted, all the related workflows need be deleted
- [x] when a workflow is deleted, all the related WorkflowTask need be deleted

ychiucco:
close #156

### DIFF
--- a/tests/test_unit_db_models.py
+++ b/tests/test_unit_db_models.py
@@ -103,67 +103,44 @@ async def test_cascade_delete_workflow(
 ):
     """
     GIVEN a Workflow
-    WHEN the Workflow is deleted via API
+    WHEN the Workflow is deleted
     THEN all the related WorkflowTask are deleted
     """
     async with MockCurrentUser(persist=True) as user:
-        project = await project_factory(user)
-        workflow = {
-            "name": "My Workflow",
-            "project_id": project.id,
-        }
-        res = await client.post(
-            "api/v1/workflow/",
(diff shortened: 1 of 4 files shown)


jacopo-exact (changes requested):
There are a few issues.

1. The file in which you wrote the test is logically orthogonal to what you are testing. This test rather belongs to the `test_db_models.py` file. You may consider renaming `test_db_models.py -> test_unit_db_models.py`, as it is a unit test suite.

2. the tests you wrote are not unit test but rather integration tests, as they make use of the API endpoints to act on the database. What we want here is to test the low level routines that interact with the database, without any mediation by other logic. You should thus use low level calls to create the data structures and then test the delete explicitly.

tests/test_unit_db_models.py
```diff
@@ -96,3 +96,97 @@ async def test_task_workflow_association(
...
+        await db.refresh(project)
+        await db.refresh(workflow1)
+        await db.refresh(workflow2)
+
+        before_delete_wf_ids = [wf.id for wf in project.workflow_list]
+
+        await db.delete(project)
+        await db.commit()
+
+        assert not await db.get(Project, project_id)
```
jacopo-exact, line 183: <3

jacopo-exact (approved):

PR #170, opened 2022-10-21
Structure of the Workflow API

ychiucco:
ref #141

### DIFF
--- a/fractal_server/app/api/v1/workflow.py
+++ b/fractal_server/app/api/v1/workflow.py
@@ -23,6 +23,8 @@ from ...models import Workflow
 from ...models import WorkflowCreate
 from ...models import WorkflowRead
 from ...models import WorkflowTaskCreate
+from ...models import WorkflowTaskRead
+from ...models import WorkflowUpdate
 from ...security import current_active_user
 from ...security import User
 from .project import get_project_check_owner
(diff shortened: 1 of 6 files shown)


fractal_server/app/api/v1/workflow.py
```diff
@@ -83,3 +84,33 @@ async def delete_workflow(
...
     return Response(status_code=status.HTTP_204_NO_CONTENT)
+
+
+@router.get("/{_id}", response_model=WorkflowRead)
+async def get_workflow(
+    _id: int,
+    user: User = Depends(current_active_user),
+    db: AsyncSession = Depends(get_db),
+):
+    workflow = db.get(Workflow, _id)
```
jacopo-exact, line 95: missing access control

fractal_server/app/api/v1/workflow.py
```diff
@@ -83,3 +84,33 @@ async def delete_workflow(
...
+        user_id=user.id,
+        db=db,
+    )
+    return workflow
+
+
+@router.patch("{_id}/add-task/")
+async def add_task_to_workflow(
+    _id: int,
+    new_task: Task,
```
jacopo-exact, line 112: needs be WorkflowTaskCreate (TBD)

jacopo-exact (changes requested):

fractal_server/app/schemas/workflow.py
```diff
@@ -1,11 +1,24 @@
...
+
+class _WorkflowTaskBase(SQLModel):
+    workflow_id: Optional[int]
+    task_id: Optional[int]
+    order: Optional[int]
+    args: Dict[str, Any]
+
+
+class WorkflowTaskCreate(_WorkflowTaskBase):
+    pass
```
jacopo-exact, line 21: `WorkflowTaskCreate.task_id` needs be mandatory

fractal_server/app/api/v1/workflow.py
```diff
@@ -83,3 +86,55 @@ async def delete_workflow(
...
+    # check autorization
+    await get_project_check_owner(
+        project_id=workflow.project_id,
+        user_id=user.id,
+        db=db,
+    )
+    return workflow
+
+
+@router.patch("{_id}/add-task/", response_model=WorkflowTaskRead)
```
jacopo-exact, line 111: sorry, this is really my bad, let's change this to POST. It is true that we are modifying the workflow, but we are most of all creating a WorkflowTask (return 201).

Also, in this case it is better to return the whole Workflow, as if we insert in the middle (`order != len(Workflow.task_list)`) it would be difficult for the clients to use the information of the task alone, as all the following tasks would change index.

fractal_server/app/schemas/workflow.py
```diff
@@ -1,11 +1,35 @@
...
+
+class _WorkflowTaskBase(SQLModel):
+    workflow_id: Optional[int]
+    task_id: Optional[int]
+    order: Optional[int]
+    args: Dict[str, Any]
+
+
+class WorkflowTaskCreate(_WorkflowTaskBase):
+    workflow_id: int
```
jacopo-exact, line 27: it is better that we leave workflow_id as optional, only task_id is really mandatory.
workflow_id is filled automatically when we add the `WorkflowTask` to a `Workflow.task_list`

jacopo-exact (changes requested):

jacopo-exact (approved):

PR #173, opened 2022-10-21
last endpoints for Workflow

ychiucco:


fractal_server/app/api/v1/workflow.py
```diff
@@ -65,7 +65,7 @@ async def create_workflow(
     return db_workflow
 
 
-@router.delete("/{_id}", status_code=status.HTTP_204_NO_CONTENT)
+@router.delete("/{_id}/", status_code=status.HTTP_204_NO_CONTENT)
```
jacopo-exact, line 68: please remove trailing '/' after path parameters.
The slash should only appear after a constant string, such as "/{_id}/add-task/", as this way we avoid a few name clashes among endpoints.

I.e.,
`/{_id}/` and `/do-something/` both match the same pattern `/*/`, but `/{_id}` and `/do-something/` do not

fractal_server/app/api/v1/workflow.py
```diff
@@ -87,13 +87,14 @@ async def delete_workflow(
     return Response(status_code=status.HTTP_204_NO_CONTENT)
 
 
-@router.get("/{_id}", response_model=WorkflowRead)
+@router.get("/{_id}/", response_model=WorkflowRead)
```
jacopo-exact, line 90: idem

fractal_server/app/api/v1/workflow.py
```diff
@@ -107,33 +108,86 @@ async def get_workflow(
...
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail="Workflow not found"
+        )
+    # check autorization
+    await get_project_check_owner(
+        project_id=workflow.project_id,
+        user_id=user.id,
+        db=db,
+    )
+    await workflow.insert_task(new_task.task_id, db=db)
```
jacopo-exact, line 134: the `order` parameter is not handled

fractal_server/app/api/v1/workflow.py
```diff
@@ -107,33 +108,86 @@ async def get_workflow(
...
+
+    db.merge(workflow)
```
jacopo-exact, line 136: not sure this is really needed. `workflow` should already be in the session

fractal_server/app/api/v1/workflow.py
```diff
@@ -107,33 +108,86 @@ async def get_workflow(
...
+    db.commit()
+    await db.refresh(workflow)
+
+    return workflow
 
 
 @router.delete(
-    "{_id}/rm-task/{task_id}", status_code=status.HTTP_204_NO_CONTENT
+    "/{_id}/rm-task/{task_id}/", status_code=status.HTTP_204_NO_CONTENT
```
jacopo-exact, line 144: see above.
let's call it `workflow_task_id` instead of `task_id`, to make it clear that we are selecting the task from the task_list by primary key.

fractal_server/app/models/workflow.py
```diff
@@ -94,3 +94,24 @@ async def insert_task(
...
+            if wftask.task.id == task_id:
+                j = i + 1
+                for _wftask in self.task_list[j:]:
+                    _wftask.order -= 1
+                    await db.merge(_wftask)
+                await db.delete(wftask)
+                self.task_list.remove(wftask)
+                await db.commit()
+                break
+        return
```
jacopo-exact, line 117: deletion should always take place by WorkflowTask.id, which is the primary key and can thus be deleted in a one liner. please remove this method

fractal_server/app/api/v1/workflow.py
```diff
@@ -107,33 +108,86 @@ async def get_workflow(
...
+        )
+    # check autorization
+    await get_project_check_owner(
+        project_id=workflow.project_id,
+        user_id=user.id,
+        db=db,
+    )
+    await workflow.remove_task(task_id, db=db)
+    await db.merge(workflow)
+    await db.commit()
```
jacopo-exact, line 166: we should simply delete the `WorkflowTask` object

fractal_server/app/api/v1/workflow.py
```diff
@@ -107,33 +108,86 @@ async def get_workflow(
...
+    return
 
 
-@router.patch("{_id}", response_model=WorkflowRead)
+@router.patch("/{_id}/", response_model=WorkflowRead)
```
jacopo-exact, line 170: see above

tests/test_workflow_api.py
```diff
@@ -79,14 +82,208 @@ async def test_workflow_delete(
...
 
         res = await client.post(
-            "api/v1/workflow/",
+            "/api/v1/workflow/",
             json=workflow,
         )
         wf_id = res.json()["id"]
 
-        res = await client.delete(f"api/v1/workflow/{wf_id}")
+        res = await client.delete(f"/api/v1/workflow/{wf_id}/")
```
jacopo-exact, line 90: it was correct before

tests/test_workflow_api.py
```diff
@@ -79,14 +82,208 @@ async def test_workflow_delete(
...
+    GIVEN a Workflow in the db
+    WHEN the get endpoint is called
+    THEN the Workflow is returned
+    """
+    async with MockCurrentUser(persist=True) as user:
+        project = await project_factory(user)
+        p_id = project.id
+        WF_ID = 1
+        workflow = {
+            "id": WF_ID,
```
jacopo-exact, line 111: the user never passes the primary key.
in this case the server is silently discarding it anyway, but please remove it

tests/test_workflow_api.py
```diff
@@ -79,14 +82,208 @@ async def test_workflow_delete(
...
+            "name": "My Workflow",
+            "project_id": p_id,
+            "task_list": [],
+        }
+        res = await client.post(
+            "/api/v1/workflow/",
+            json=workflow,
+        )
+
+        res = await client.get(f"/api/v1/workflow/{WF_ID}/")
```
jacopo-exact, line 121: the id needs be read from the result of the POST call.
remove trailing /

tests/test_workflow_api.py
```diff
@@ -79,14 +82,208 @@ async def test_workflow_delete(
...
+        assert res.status_code == 200
+        assert res.json() == workflow
+
+
+async def test_post_newtask(
+    db, client, MockCurrentUser, project_factory, task_factory
+):
+    """
+    GIVEN a Workflow with a list of Tasks
+    WHEN the POST endpoint is called
```
jacopo-exact, line 131: specify the endpoint you're calling. e.g., "the endpoint to add a task to the task list is called"

tests/test_workflow_api.py
```diff
@@ -79,14 +82,208 @@ async def test_workflow_delete(
...
+    THEN the new Task is inserted in Workflow.task_list
+    """
+    async with MockCurrentUser(persist=True) as user:
+        project = await project_factory(user)
+        WF_ID = 1
+        workflow = {
+            "id": WF_ID,
```
jacopo-exact, line 138: the id is the server's concern. you must read it from the response of the POST call.

tests/test_workflow_api.py
```diff
@@ -79,14 +82,208 @@ async def test_workflow_delete(
...
+
+        assert len(workflow.task_list) == 2
+        assert workflow.task_list[0].task == t0
+        assert workflow.task_list[1].task == t1
+
+        t2 = await task_factory()
+        new_task = {
+            "task_id": t2.id,
+        }
+
```
jacopo-exact, line 163: this will append the task at the end of the task list. please also test the case in which we insert the task in the middle

tests/test_workflow_api.py
```diff
@@ -79,14 +82,208 @@ async def test_workflow_delete(
...
+        )
+        assert res.status_code == 201
+        debug(res.json())
+        workflow = WorkflowRead(**res.json())
+
+        assert len(workflow.task_list) == 3
+        assert workflow.task_list[0].task == TaskRead(**t0.__dict__)
+        assert workflow.task_list[1].task == TaskRead(**t1.__dict__)
+        assert workflow.task_list[2].task == TaskRead(**t2.__dict__)
+
```
jacopo-exact, line 176: please test on the raw data, as the model could silently add defaults and give false positives

jacopo-exact (changes requested):
i still need to review the last three tests

tests/test_workflow_api.py
```diff
@@ -90,3 +93,195 @@ async def test_workflow_delete(
...
+            json=new_task,
+        )
+        assert res.status_code == 201
+        debug(res.json())
+        workflow = WorkflowRead(**res.json())
+
+        assert len(workflow.task_list) == 3
+        assert workflow.task_list[0].task == TaskRead(**t0.__dict__)
+        assert workflow.task_list[1].task == TaskRead(**t1.__dict__)
+        assert workflow.task_list[2].task == TaskRead(**t2.__dict__)
```
jacopo-exact, line 175: - [ ] use public method `Task.dict()` instead of private  `__dict__`
- [ ] add test in which you insert a task in the middle instead of appending it at the end of the task_list.

tests/test_workflow_api.py
```diff
@@ -90,3 +93,195 @@ async def test_workflow_delete(
...
+    WHEN the endpoint to GET all the Workflows associated
+        to that Project is called
+    THEN the list of all its Workflows is returned
+    """
+    async with MockCurrentUser(persist=True) as user:
+        project = await project_factory(user)
+        other_project = await project_factory(user)
+        workflow1 = {"name": "WF1", "project_id": project.id}
+        workflow2 = {"name": "WF2", "project_id": project.id}
+        workflow3 = {"name": "WF3", "project_id": other_project.id}
```
jacopo-exact, line 243: molto bene!

jacopo-exact (changes requested):
very minor changes

jacopo-exact (approved):

PR #182, opened 2022-10-25
Use parsl executors in v1.x

### LINKED ISSUE #82: Handle executors in v1.x
executor is a configuration detail of the single installation of fractal server and as such should not be part of the task model.

cc @tcompa

tcompa:
Close #82 

Note: this only concerns the `_parsl` backend (not `_process`).


fractal_server/app/models/task.py
```diff
@@ -24,3 +24,10 @@ def parallelization_level(self) -> Optional[str]:
     @property
     def is_parallel(self):
         return bool(self.parallelization_level)
+
+    @property
+    def executor(self) -> Optional[str]:
+        try:
+            return self.default_args["executor"]
+        except KeyError:
+            return None
```
jacopo-exact, line 33: i would very much like to avoid having parsl specific stuff in the task definition. For now we can keep it like this, but we should work towards adding objective constraints in the tasks and move the selection of a specific executor to the backends or as close as possible to the backends.

E.g.,
* Task says "I needs 4GB memory"
* backend says "the executor that fits your requirements is `hi-mem-blabla`"

tests/test_parsl_executors.py
```diff
@@ -0,0 +1,155 @@
+"""
```
jacopo-exact, line 1: I would move these tests to the `tests/test_backend_parsl.py` or rename this test something along the lines of `tests/test_backend_parsl_executors.py`, so as to have clear that this stuff logically belongs to the runner backend subsystem and in particular the parsl backend.

tests/test_parsl_executors.py
```diff
@@ -0,0 +1,155 @@
...
+    )
+    close_job_logger(logger)
+    debug(out)
+    assert "dummy" in out.metadata
+    assert "dummy" in out.metadata
+    assert out.metadata["history"] == [
+        tk.name,
+        tk.name,
+        f"{tp.name}: [0, 1, 2]",
+    ]
```
jacopo-exact, line 99: Should there be an assertion that the task has run in the correct parsl executor?

jacopo-exact (changes requested):

fractal_server/app/models/task.py
tcompa: OK, let's keep this in mind

tests/test_parsl_executors.py
tcompa: I'm going with `test_backend_parsl_executors`, because it is reasonably different from `test_backend_parsl.py` (the new one uses actual objects instead of mocks, it uses the db, it uses `process_worfklow`).

tests/test_parsl_executors.py
tcompa: Agreed, that would be the actual test. However:
* This information ("I ran task on that executor") is AFAIK only available through the logs written by fractal-server, or through parsl inner objects which are not easy to access.
* Enforcing strict rules for the logs does not seem to be the a relevant feature to be tested.

jacopo-exact (approved):

PR #183, opened 2022-10-25
Improve test of non-trivial workflows

### LINKED ISSUE #165: Add test for generic non-trivial workflow (with parallel/global tasks)

tcompa:
Close #165 

EDIT: almost all the required tests were already available in `test_runner_backends`. We still lack a full test (using `submit_workflow` instead of `process_workflow`), but that's out of scope for this PR.

In the end this PR only adds a more stringent test on the history.


tcompa:
Notice that I broke another test:
```python traceback
FAILED tests/test_backend_process.py::test_recursive_task_submission_inductive_step - AssertionError: assert {'dummy': 'du...x': [0, 1, 2]} == {'dummy': 'du...x': [0, 1, 2]}
``

Issue #186, opened 2022-10-26
use tmp_path for test db


jacopo-exact:
this requires that we cook up a fixture with `scope="session" ` temporary paths

PR #188, opened 2022-10-26
Task collection

### LINKED ISSUE #180: task collection 1.x from pypi
Preliminary requirement: we need a shared folder that is readable and executable by the fractal user, let it be `FRACTAL_ROOT_DIR`

# from PyPI

in a subprocess...

1. create dedicated venv in `FRACTAL_ROOT_DIR/<pkg>_<version>`
2. install package in venv
3. extract manifest by `/venv/python -c "import <pkg>; print(pkg.__manifest__)"`
4. load all tasks exposed in manifest

## default behaviour

```bash
cd $FRACTAL_ROOT_DIR
PKG_DIR=$PKG_NAME_$PKG_VERSION
mkdir $PKG_DIR
cd $PKG_DIR
python -m venv venv (or similar for conda and the like)
source /venv/bin/activate (or similar)
(venv) pip install $PKG_NAME==$PKG_VERSION
(venv) python -c "import $PKG_NAME; print($PKG_NAME.__manifest__)" > /tmp/fractal/pkg_name_version.manifest.json
```
we read the output of the last command and access the manifest from within the server

the resulting tasks will have
- source = pypi:fractal-tasks-core==0.2.3
- command =  /path/to/venv/python path/to/module.py

## custom behaviour (TODO)
We keep the folder structure with the following pseudo-script

```bash
cd $FRACTAL_ROOT_DIR
PKG_DIR=$PKG_NAME_$PKG_VERSION
mkdir $PKG_DIR
cd $PKG_DIR
```

the user can provide a custom initialisation / installation script, which needs to output
1. how to call the dedicated python interpreter
2. the manifest

# from `FRACTAL_ROOT_DIR/<userdir>/<taskdir>/`

TODO

jacopo-exact:
close #180, #186


tests/fixtures_server.py
```diff
@@ -33,19 +34,25 @@
...
 
 
-def get_patched_settings():
+def get_patched_settings(temp_path: Path):
     settings = Settings()
     settings.JWT_SECRET_KEY = "secret_key"
     settings.DEPLOYMENT_TYPE = "development"
     settings.DB_ENGINE = "sqlite"
-    settings.SQLITE_PATH = "_test.db?mode=memory&cache=shared"
+    settings.SQLITE_PATH = (
```
tcompa, line 42: Sorry, I cannot find `temp_path` definition. Is it custom or from pytest?

fractal_server/config.py
```diff
@@ -137,14 +141,16 @@ def DATABASE_URL(self):
...
+                raise ValueError("SQLITE_PATH path cannot be None")
             return self.DATABASE_URL.replace("aiosqlite", "pysqlite")
         elif self.DB_ENGINE == "postgres":
             return self.DATABASE_URL.replace("asyncpg", "psycopg2")
 
     ###########################################################################
     # FRACTAL SPECIFIC
     ###########################################################################
-    # FRACTAL_ROOT: Path = Path("FRACTAL_ROOT")
+    FRACTAL_ROOT: Path = _DEVNULL
```
tcompa, line 153: Any special reason for the difference between `FRACTAL_ROOT` (default: `_DEVNULL`) and `SQLITE_PATH` (default: `None`)?

fractal_server/app/schemas/manifest.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class TaskManifestV1(_TaskManifestBase):
+    pass
+
+
+class ManifestV1(_ManifestBase):
+    task_list: List[TaskManifestV1]
+
+    @validator("manifest_version")
```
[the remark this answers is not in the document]
tcompa: Another side-comment (not needed for this PR)
It would be useful to expose the validator as a simple command-line tool, through a simple wrapper, for the user to validate their custom manifest (for custom tasks). Something like
```
validate_manifest.py /path/to/my/manifest
```

fractal_server/app/db/__init__.py
```diff
@@ -7,49 +7,63 @@
...
-
-async def get_db() -> AsyncGenerator[AsyncSession, None]:
-    async with async_session_maker() as async_session:
-        yield async_session
-
-
-def get_sync_db() -> Generator[DBSyncSession, None, None]:
-    with sync_session_maker() as sync_session:
-        yield sync_session
+class DB:
```
tcompa, line 18: If it can be explained in a sentence: what's the reason for encapsulating this code block in the new `DB` class?

fractal_server/config.py
jacopo-exact: No, you're right. Will simplify this by making FRACTAL_ROOT an optional path.

tests/fixtures_server.py
jacopo-exact: notice that `get_patched_settings(temp_path)` is just a function, not a fixutre. Thus `temp_path` is just a parameter name.

Indeed `tmp_path` is a standard fixture of pytest.

fractal_server/app/schemas/manifest.py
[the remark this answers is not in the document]
jacopo-exact: > Another side-comment (not needed for this PR) It would be useful to expose the validator as a simple command-line tool, through a simple wrapper, for the user to validate their custom manifest (for custom tasks). 

Indeed, that's why I decided to make a pydantic schema for it. Keep in mind that when the refactoring will be over we could either mirror or move `app/schemas` in `fractal_common`. At that point it is trivial to add the validation utility to the client cli.

PR #221, opened 2022-11-08
Move `executor` and `parallelization_level` to `task.meta`

### LINKED ISSUE #216: Extract `executor` and `parallelization_level` from `meta`, rather than from `default_args`
In recent versions of the manifest schema, the `meta` field of a task may looks like
```python
"meta": {"executor": "cpu-low"}
```
but we are still looking for `executor` in `default_args`.

tcompa:
Close #216

### DIFF
--- a/fractal_server/app/models/task.py
+++ b/fractal_server/app/models/task.py
@@ -12,12 +12,14 @@ from ..schemas.task import _TaskBase
 class Task(_TaskBase, table=True):  # type: ignore
     id: Optional[int] = Field(default=None, primary_key=True)
     project_id: Optional[int] = Field(foreign_key="project.id")
-    default_args: Optional[Dict[str, Any]] = Field(sa_column=Column(JSON))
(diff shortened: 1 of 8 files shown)


fractal_server/app/models/task.py
```diff
@@ -12,13 +12,15 @@
 class Task(_TaskBase, table=True):  # type: ignore
     id: Optional[int] = Field(default=None, primary_key=True)
     project_id: Optional[int] = Field(foreign_key="project.id")
+    # FIXME: is default={} needed for default_args and meta?
     default_args: Dict[str, Any] = Field(sa_column=Column(JSON), default={})
```
tcompa, line 16: Questions:
* Should both `default_args` and `meta` be `Optional`?
* Is `default={}` needed?

tcompa: @jacopo-exact

jacopo-exact: probably better to have them optional, allow none and check a posteriori

jacopo-exact (approved):

Issue #225, opened 2022-11-09
Fail for `add-task` if `task_id` does not exist


tcompa:
After creating a workflow (via the client), issuing `fractal workflow add-task 1 1` gives a non-informative error:
```
$ fractal workflow add-task 1 1

DEBUG:asyncio:Using selector: EpollSelector
DEBUG:root:Namespace(batch=False, cmd='workflow', id='1', json=False, order=None, password=None, slurm_user=None, subcmd='add-task', task_id='1', user=None, v=0)
DEBUG:httpx._client:HTTP Request: POST http://localhost:8002/api/v1/workflow/1/add-task/ "HTTP/1.1 500 Internal Server Error"
DEBUG:root:500
ERROR:root:Server returned 500
ERROR:root:Original request: POST http://localhost:8002/api/v1/workflow/1/add-task/
ERROR:root:Original payload: {"workflow_id": null, "task_id": 1, "order": null, "args": null}
ERROR:root:Server error message: {}

ERROR:root:Terminating.
```

The error in the server logs read
```python traceback
INFO:     127.0.0.1:36730 - "POST /api/v1/workflow/1/add-task/ HTTP/1.1" 500 Internal Server Error
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/data/homes/fractal/.conda/envs/fractal-server-v1/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 404, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/data/homes/fractal/.conda/envs/fractal-server-v1/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/data/homes/fractal/.conda/envs/fractal-server-v1/lib/python3.8/site-packages/fastapi/applications.py", line 269, in __call__
    await super().__call__(scope, receive, send)
  File "/data/homes/fractal/.conda/envs/fractal-server-v1/lib/python3.8/site-packages/starlette/applications.py", line 124, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/data/homes/fractal/.conda/envs/fractal-server-v1/lib/python3.8/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/data/homes/fractal/.conda/envs/fractal-server-v1/lib/python3.8/site-packages/starlette/middleware/errors.py", line 162, in __call__
[...]

PR #229, opened 2022-11-09
test for foreign key violation

jacopo-exact:
ref #225


jacopo-exact:
issue is with sqlite's handling of foreign keys. it does not seem to be immediately solvable with sqlite.

tcompa:
For the record, another example where postgres catches an error (in tests) that sqlite doesn't:
https://github.com/fractal-analytics-platform/fractal-server/pull/372#issuecomment-1361337419

PR #237, opened 2022-11-10
Sanitize component to build JSON-file names

### LINKED ISSUE #235: Sanitize `component` before using it to name args JSON file, in runner

### LINKED ISSUE #238: `component` must be a string, also in dummy tasks and tests
(this came up during #237)

tcompa:
Close #238 
Close #235


jacopo-exact (approved):
Let us consider if `str` is the right type for components or we are hitting limitations that could have consequences down the road.

PR #252, opened 2022-11-11
Provide meaningful error for non valid manifest

tcompa:
MISSING:
How to test that the task collection failed, due to invalid manifest? Loading tasks from an invalid manifest raises a ValidationError, that is handled in the server as 422, but this happens in background.

TL;DR how to test for a failure in `_background_collect_pip`?

@jacopo-exact @mfranzon


tcompa:
The useful bits of this PR are already included in https://github.com/fractal-analytics-platform/fractal-server/pull/256, the rest is useless. Closing.

Issue #215, opened 2022-11-07
When collecting tasks with `version=None`, fill `source` with the actual version


tcompa:
Right now we have
```python
'source': 'pip:fractal-tasks-core==None',
```

My guess is that we want to show the actual version that was installed.

jacopo-exact:
do we want to give the possibility to leave the version blank and update the package? kind of like tasks-latest...

tcompa:
> do we want to give the possibility to leave the version blank and update the package? kind of like tasks-latest...

My take:
* Yes, you can call the endpoint without specifying any version. In that case, the install command is just `pip install package`.
* The task/env folders should all have a version in their name (i.e. `fractal-tasks-core-1.2.3` is valid, `fractal-tasks-core` or `fractal-tasks-core-latest` are not valid). Having some non-standard name would be confusing, when more than one version is available in Fractal. There are (at least) two ways to achieve this:
    1. Install the package, find the current version, rename the env folder.
    2. Find out the latest available version, then proceed in the standard way with that version specified. This requires a preliminary `pip` call. `pip search` would do it, but it's not available any more. Now we may use something like `pip install package --dry-run  --report /tmp/report.json`, and parse the output to extract the latest version.

jacopo-exact:
Ok.
This can be achieved by splitting installation in download + local install.

```bash
# download without dependencies
pip download --no-deps <pkg>
# install the local wheel
pip install <local pkg>
```

incidentally, this allows us to inspect the package before installation and read the manifest before installation, cf. #250

PR #256, opened 2022-11-11
215 consider when version is not provided

jacopo-exact:
fill package version information in source if needed.

add test

ref #215


tests/test_task_api.py
```diff
@@ -22,6 +27,41 @@ async def test_task_get_list(db, client, task_factory, MockCurrentUser):
...
+        task_pkg=task_pkg, user="test_bg_collection"
+    )
+    collection_status = TaskCollectStatus(
+        status="pending",
+        venv_path=venv_path,
+        package=task_pkg.package
+        # **task_pkg.dict(),
+    )
+    # replacing with path because of non-serializable Path
+    collection_status_dict = collection_status.dict()
```
tcompa, line 47: Can we use `sanitised_dict` here?

tests/test_task_api.py
```diff
@@ -22,6 +27,41 @@ async def test_task_get_list(db, client, task_factory, MockCurrentUser):
...
+    """
+    task_pkg = _TaskCollectPip(package=dummy_task_package.as_posix())
+    venv_path = create_package_dir_pip(
+        task_pkg=task_pkg, user="test_bg_collection"
+    )
+    collection_status = TaskCollectStatus(
+        status="pending",
+        venv_path=venv_path,
+        package=task_pkg.package
+        # **task_pkg.dict(),
```
tcompa, line 44: Is this comment relevant?

fractal_server/app/api/v1/task.py
```diff
@@ -94,51 +152,67 @@ async def collect_tasks_pip(
...
-    settings = Inject(get_settings)
 
-    return TaskCollectResult(
-        **task_pkg.dict(),
-        venv_path=venv_path.relative_to(settings.FRACTAL_ROOT),
-        info=(
-            "Collecting tasks in the background. "
-            "GET /task/collect/{venv_path} to query collection status"
-        ),
+    logger.info("collection endpiont: returning state")
```
tcompa, line 179: typo

tcompa (approved):

tcompa:
One last comment, for my understanding.
I added `import time; time.sleep(10)` within `_background_collect_pip` (for testing), and then ran `poetry run pytest tests/test_task_api.py::test_collection_api`. I was expecting this block to fail
```python
       res = await client.post(
            f"{PREFIX}/collect/pip/?public=false", json=task_collection
        )
        debug(res.json())
        assert res.status_code == 201
        assert res.json()["data"]["status"] == "pending"

        state = res.json()
        data = state["data"]
        assert "fractal_tasks_dummy" in data["venv_path"]
        venv_path = Path(data["venv_path"])

        res = await client.get(f"{PREFIX}/collect/{state['id']}")
        debug(res.json())
        assert res.status_code == 200
        status = res.json()
        data = status["data"]

        assert data["status"] == "OK"
```
because the background task should have taken longer than the time to get to the OK assert statement.

What I observed is that the test passed, but simply took 10 seconds longer. I don't understand which call was blocking, and why the 10 seconds were not being waited in background.

(nothing to change, most likely, but I just want to understand better)

tests/test_task_api.py
jacopo-exact: yes, thanks

tests/test_task_api.py
jacopo-exact: no, thanks

tests/test_task_api.py
```diff
@@ -46,16 +86,18 @@ async def test_collection_api(client, dummy_task_package, MockCurrentUser):
...
+        data = state["data"]
         assert "fractal_tasks_dummy" in data["venv_path"]
         venv_path = Path(data["venv_path"])
 
-        res = await client.get(f"{PREFIX}/collect/{venv_path}")
+        res = await client.get(f"{PREFIX}/collect/{state['id']}")
         debug(res.json())
         assert res.status_code == 200
-        data = res.json()
+        status = res.json()
```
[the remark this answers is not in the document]
jacopo-exact: ok

tests/test_task_api.py
```diff
@@ -67,14 +109,39 @@ async def test_collection_api(client, dummy_task_package, MockCurrentUser):
         assert "dummy parallel" in task_names
 
         # using verbose option
-        res = await client.get(f"{PREFIX}/collect/{venv_path}?verbose=true")
+        res = await client.get(f"{PREFIX}/collect/{state['id']}?verbose=true")
         debug(res.json())
+        status = res.json()
```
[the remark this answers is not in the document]
jacopo-exact: ok

jacopo-exact:
> One last comment, for my understanding. I added `import time; time.sleep(10)` within `_background_collect_pip` (for testing), and then ran `poetry run pytest tests/test_task_api.py::test_collection_api`. I was expecting this block to fail
> 
> ```python
>        res = await client.post(
>             f"{PREFIX}/collect/pip/?public=false", json=task_collection
>         )
>         assert res.status_code == 201
>         assert res.json()["data"]["status"] == "pending"
> 
>         ...
> 
>         res = await client.get(f"{PREFIX}/collect/{state['id']}")
>         assert res.status_code == 200
>         status = res.json()
>         data = status["data"]
> 
>         assert data["status"] == "OK"
> ```
> 
> because the background task should have taken longer than the time to get to the OK assert statement.
> 
> What I observed is that the test passed, but simply took 10 seconds longer. I don't understand which call was blocking, and why the 10 seconds were not being waited in background.
> 
> (nothing to change, most likely, but I just want to understand better)

Most likely it did not fail because you used the synchronous sleep and the test runs in an event loop within a single thread. `_background_collect_pip` acquired the loop and could not yield back because of the synchronous call.

I would expect that if you repeted your experiment with an `await asyncio.sleep(10)` the test would fail.

PR #257, opened 2022-11-11
240 include version in tasks package directory server or client issue

jacopo-exact:
@tcompa fractal-analytics-platform/fractal#326 might be a client error that does not send the version number to the server. Unit tests added by this PR behave as expected.


jacopo-exact:
confirmed client error. issue is moved to client repo. let's merge this PR anyways, as there is an extra test and some config refactoring

Issue #224, opened 2022-11-09
Setup stress-test script


tcompa:
First version: fractal user submits 50 workflows.
Second version: 10 different users submit 5 workflows each.
Third version: fractal user submits a workflow with 1000 parallel tasks.

tcompa:
The first stress test is available in https://github.com/fractal-analytics-platform/fractal-demos/tree/main/Fractal_stress_test (as of commit https://github.com/fractal-analytics-platform/fractal-demos/commit/43c0833149726e7c8b3e7471756b9a9c3cc51774).
The script `run_example.sh` runs a full-workflow example (based on the tasks in the `fractal-tasks-stresstest` package, which just create a list of components and then sleep a certain amount of time for each component), through the client. The script `simulate_many_runs.sh` submits several copies of this examples.

When reaching 10-15 examples running simultaneously, the server crashes with a "Too many open files" error. I will report more details later.

tcompa:
This has been reviewed with server `1.0.0b1`, and using the slurm backend. The stress test is defined in https://github.com/fractal-analytics-platform/fractal-demos/tree/c98154801209c672d117f2b9ad8cc3789bfe012c/Fractal_stress_test, and it submits 20 workflows like this one:
* A preliminary task that prepares a list of components of length 20.
* 20 parallel tasks that do nothing for a while (200 seconds)

The slurm `config.json` file reads
```JSON
{
    "cpu-low": {
            "partition": "main",
            "mem": "1g",
            "nodes": "1",
            "ntasks-per-node": "1",
            "cpus-per-task": "1"
        }
}
```

The script correctly submits the 20 workflows (as a user), and (on UZH) some start running while others are listed in `squeue` as
```
9410755        test01     test01                     sleep_task  PENDING       0:00 7-00:00:00    1      1024M (AssocMaxJobsLimit) 
```
which is an expected behavior.

The whole stress test ended in 500 seconds, and it ran:
* 20 tasks which run in essentially no time (just preparing the metadata);
* 400 tasks that sleep for 200 seconds each.

Each one of the job folders has this kind of files, after execution:
```
$ ls artifacts/workflow_000001_job_000001/
0.args.json		1_par_12.out		1_par_17.metadiff.json	1_par_3.err	       1_par_8.args.json	   slurmpy.stdout.9410466.log		      _temp_JSH8dVkkkJrErjhoXG5Ilp12WoaXrFrk.sh
[...]

PR #266, opened 2022-11-14
Set up stress test

tcompa:
Ref #224


tcompa:
I will develop this in the fractal-demos repo, and then eventually bring back here what is relevant.

PR #267, opened 2022-11-15
 ref #230 , add executor to workflowTask model, update workflowtask 

### LINKED ISSUE #9: Forbid a task update that changes `meta["parallelization_level"]`
If a core task is designed to work e.g. at the well level, it means it will not work if it is parallelized over plates.
The manifest should then specify the right `parallelization_level` for each one of the core tasks, and this should not be overridden later.

### LINKED ISSUE #230: Allow override of some `Task.meta` parameters within `WorkflowTask`
This is necessary to modify `meta["executor"]` of a `WorkflowTask` object.

mfranzon:
closes #9 
closes #230


fractal_server/app/models/workflow.py
```diff
@@ -40,6 +40,7 @@ class Config:
     task_id: Optional[int] = Field(foreign_key="task.id")
 
     order: Optional[int]
+    executor: Optional[str] = None
```
tcompa, line 43: Ambiguity between property and attribute names.

jacopo-exact: Attention!
you are adding to the base model an attribute that is specific to a single backend. Please consider subclassing within the backend.

PR #270, opened 2022-11-15
234 better handling of task collection error when folder already exists

jacopo-exact:


fractal_server/tasks/collection.py
```diff
@@ -123,7 +154,8 @@ def create_package_dir_pip(
     package_dir = f"{task_pkg.package}{task_pkg.version or ''}"
     venv_path = settings.FRACTAL_ROOT / user / package_dir  # type: ignore
     # TODO check the access right of the venv_path and subdirs
-    venv_path.mkdir(exist_ok=False, parents=True)
+    if create:
+        venv_path.mkdir(exist_ok=False, parents=True)
```
tcompa, line 158: We'll have to make sure that this sets the correct permissions. [The default is 0o777](https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir), which is OK, but in principle it could be modified by a non-zero umask.

I think this already works, because we don't need any write permissions by users on this folder, but this is just a reminder that if something fails with permission errors we may have to come back to this.

tcompa (approved):

PR #286, opened 2022-11-18
BUGFIX Await download-package function

tcompa:


ychiucco:
Not awaited coroutines generate `RuntimeWarning`s.
If we run `pytest` as `pytest test.py -W error::RuntimeWarning`, we get a `PytestUnraisableExceptionWarning`, and the test still pass.

I thought of this solution
```python
import functools
import warnings

def checkawait(my_test):
    @functools.wraps(my_test)
    async def wrapper(*args, **kwargs):
        with pytest.warns(RuntimeWarning) as record:
            await my_test(*args, **kwargs)
            captured_warnings = record.list
            if not captured_warnings:
                warnings.warn("Harmless warning", RuntimeWarning)
        for warning in captured_warnings:
            if "never awaited" in warning.message.args[0]:
                raise Exception(warning.message.args)
    return wrapper
```

This make fail the first test but pass the second
```python
async def func():
    return 42

@checkawait
async def test_without_await():
    func()

@checkawait
async def test_with_await():
    await func()
```

Issue #284, opened 2022-11-18
In SLURM backend, set the slurm script and output/error to custom paths


tcompa:
I think that the `WORKFLOW_DIR` folder for a single-global-task workflow should look like:
```
0.args.json
0.err
0.out
0.metadiff.json
0.slurm.submit
0.slurm.err
0.slurm.out
```
this requires adding lines like:
```
#SBATCH --output=/SOME/PATH.out
#SBATCH --output=/SOME/PATH.err
```
in the sbatch file (defined here: https://github.com/sampsyo/clusterfutures/blob/master/cfut/slurm.py), and possibly also change the location of the submission script (in the ` submit_text` function).

jacopo-exact:
for parallel tasks, it is not possible to set custom per-component files unless we give up `Executor.map` altogether

tcompa:
Quick possible workaroun: can we create a folder with a simple name (`1.artifacts`) and plate out/err files in there, with arbitrary names?

tcompa:
Or, something else:
If these files go to some other arbitrary paths (the defaults are typically `jobid.err` and `jobid.out`, in **some** folder), but can we at least identify those two paths and write them into one of fractal logs?

tcompa:
Last fall-back option:
Is it "easy" to do it at least for serial tasks only? That would already be useful.

jacopo-exact:
serial tasks are almost trivial.
the problem is with parallel tasks, as the component is assigned to each slurm job only within the map function.

tcompa:
Quick update: it seems that now all files are in the correct folder, including the slurm submission/output files for parallel tasks. Great. Missing points:
- [ ] Conform to a standard filename (instead of things like `slurmpy.stdout.9410935.log` and `_temp_nJEN0ZYTWL4vhalvvkOKGwdTZaDk2cIf.sh`, I would expect `something.slurm.out`  and `something.slurm.sh`). This is not crucial, but would be nice.
- [ ] Also include the `--error` additional `SBATCH` line (in the same way as the `--output` one), pointing some `something.slurm.err`.

jacopo-exact:
working on it right now

Issue #278, opened 2022-11-17
Set ClusterFutures SLURM options


tcompa:
Each executor has specific partition/memory/CPUs.

tcompa:
Even better (and quite simple): override [slurm.submit](https://github.com/sampsyo/clusterfutures/blob/master/cfut/slurm.py#L20) so that it takes additional parameters. Note that this is also necessary for #284.

In general, we will also need to modify the `submit_text` function, in view of https://github.com/fractal-analytics-platform/fractal-server/issues/279. Thus essentially we should replace the whole `slurm.py` file with something like `slurm_fractal.py`, and use that one in the cfut `__init__`, unless I'm missing something.

tcompa:
See also https://github.com/fractal-analytics-platform/fractal-server/issues/284, AKA let's store the sbatch script in the `WORKFLOW_DIR`.

Issue #169, opened 2022-10-21
Allow Fractal to be able to send jobs to different slurm clusters within a federation


gusqgm:
From the resource infrastructure point of view, running things with the current Drogon architecture means making luigi send jobs to either a dedicated CPU VM or to a physical workstation that has a Quadro RTX 8000 for processing segmentation with pre-trained neural networks.From the resource infrastructure point of view, running things with the current Drogon architecture means making luigi send jobs to either a dedicated CPU VM or to a physical workstation that has a Quadro RTX 8000 for processing segmentation with pre-trained neural networks.

From the perspective of the lab, it would be useful to expand our minislurm cluster in order to add the gpu workstation to it as a new node so we could run with the same resource capacity as Drogon currently.

One logical step is then to make the large FMI cluster and our minislurm part of the same federation, so that jobs can be sent either to our local or to the institute wide slurm depending on some basic parameters (e.g. queue size, resource demands, etc).

I have been checking the [slurm documentation](https://slurm.schedmd.com/federation.html) and the [slurm.conf docs](https://slurm.schedmd.com/slurm.conf.html#index) it looks like creating a federation of clusters should conceptually not be so difficult as long as ideally all of the related clusters run the same version of slurm (which will be our case very soon as we are currently updating the clusters). Each cluster keeps its own database. Creation of a federaion is done with the `sacctmgr ` command and in practice Fractal would need to add a `-M,--clusters=<cluster_list>` and the `--cluster-constraint=[!]<constraint_list>` options. 

This would be very useful for us, as we would then have the flexibility to run Fractal within our own slurm resources or be able to outsorce larger jobs to the large cluster without the need of having two independent Fractal installations. 

I do see however that this can be thought right now as a rather more specific implementation case, so here is the important question: How much work would this mean for the fractal-server? Is there anything in the current architecture that would prevent us from having this road available?

tcompa:
Are those options (`-M,--clusters=<cluster_list>` and `--cluster-constraint=[!]<constraint_list>`) related to `sbatch` or to `sacctmgr`?

Briefly: whatever you can do with a `sbatch` script should be "easily" doable within fractal:
* In the current version, the `sbatch` script is generated by parsl, which lets you add any additional configuration (see parameters `scheduler_options` in [SlurmProvider](https://parsl.readthedocs.io/en/stable/stubs/parsl.providers.SlurmProvider.html)).
[...]

PR #290, opened 2022-11-18
Slurmexecutor

jacopo-exact:


fractal_server/app/runner/_process/__init__.py
```diff
@@ -173,3 +59,29 @@ async def process_workflow(
...
+async def process_workflow(
+    *,
+    workflow: Workflow,
+    input_paths: List[Path],
+    output_path: Path,
+    input_metadata: Dict[str, Any],
+    logger_name: str,
+    workflow_dir: Path,
+    username: str = None,
+    worker_init: Optional[
```
tcompa, line 73: We'll need to look back at this also in the SLURM backend, since we may want to include additional lines for a task. Nothing urgent, let's do this once we have relevant use cases.

fractal_server/app/runner/_slurm/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+from .._common import recursive_task_submission
+from ..common import async_wrap
+from ..common import TaskParameters
+from .executor import FractalSlurmExecutor
+
+
+"""
+Slurm Bakend
+
+This backend runs fractal workflows as separate processes using a python
```
tcompa, line 17: Out-of-date docstring

fractal_server/app/runner/_slurm/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+def _process_workflow(
+    *,
+    workflow: Workflow,
+    input_paths: List[Path],
+    output_path: Path,
+    input_metadata: Dict[str, Any],
+    logger_name: str,
+    workflow_dir: Path,
+    username: str = None,
+    worker_init: Optional[
```
tcompa, line 34: The same comment in the process runner was meant to be here. Possibly we'll have to look at this variable again, but let's think about it when appropriate.

(nothing to do here)

fractal_server/app/runner/_slurm/executor.py
```diff
@@ -0,0 +1,228 @@
...
+from ....utils import set_logger
+
+
+def get_slurm_script_dir() -> Path:
+    settings = Inject(get_settings)
+    script_dir = settings.RUNNER_ROOT_DIR  # type: ignore
+    return script_dir  # type: ignore
+
+
+def get_stdout_filename(arg: str = "%j") -> Path:
```
tcompa, line 33: See #284.
(this can be done later, it's FYI)

fractal_server/app/runner/_slurm/executor.py
```diff
@@ -0,0 +1,228 @@
...
+    Write batch script
+
+    Returns:
+        batch_script_path:
+            The path to the batch script
+    """
+    if not script_dir:
+        script_dir = get_slurm_script_dir()
+
+    batch_script_path = script_dir / f"_temp_{random_string()}.sh"
```
tcompa, line 58: As above, I think we should move this to a specific folder #284.
Not urgent, can be done in future "cleanup" PRs.

fractal_server/app/runner/_slurm/executor.py
```diff
@@ -0,0 +1,228 @@
...
+    jobid = output.stdout
+    # NOTE after debugging this can be uncommented
+    # filename.unlink()
+    return int(jobid)
+
+
+def compose_sbatch_script(
+    cmdline: List[str],
+    # NOTE: In SLURM, `%j` is the placeholder for the job_id.
+    outpat: Optional[Path] = None,
```
tcompa, line 110: typo (outpat->output or outpat->outpath) throughout the function

fractal_server/app/runner/_slurm/executor.py
```diff
@@ -0,0 +1,228 @@
...
+            destination of temporary script files
+
+    Returns:
+        jobid:
+            integer job id as returned by `sbatch` submission
+    """
+    filename = write_batch_script(
+        sbatch_script=sbatch_script, script_dir=script_dir
+    )
+    submit_command = f"sbatch --parsable {filename}"
```
tcompa, line 89: (nothing to do here)

For my own information, `--parsable` man info reads
```
       --parsable
              Outputs only the job id number and the cluster name if  present.
              The  values  are  separated by a semicolon. Errors will still be
              displayed.
```

fractal_server/app/runner/_slurm/executor.py
```diff
@@ -0,0 +1,228 @@
...
+    try:
+        output = subprocess.run(  # nosec
+            full_cmd, capture_output=True, check=True
+        )
+    except subprocess.CalledProcessError as e:
+        logger = set_logger(logger_name="slurm_runner")
+        logger.error(e.stderr)
+        raise e
+    jobid = output.stdout
+    # NOTE after debugging this can be uncommented
```
tcompa, line 102: Is this comment still relevant?

fractal_server/app/runner/_slurm/executor.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+    try:
+        output = subprocess.run(  # nosec
+            full_cmd, capture_output=True, check=True
+        )
+    except subprocess.CalledProcessError as e:
+        logger = set_logger(logger_name="slurm_runner")
+        logger.error(e.stderr)
+        raise e
+    jobid = output.stdout
```
tcompa, line 101: Can we also cover the case where `output="123; clustername"`? The man page says that also the cluster name is printed, if present. And according to #169 this cluster name _could_ be present.

fractal_server/app/runner/_slurm/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Returns:
+    output_dataset_metadata (Dict):
+        the output metadata
+    """
+
+    with FractalSlurmExecutor(
+        debug=True,
+        keep_logs=True,
+        additional_setup_lines=("#SBATCH --partition=main",),
```
tcompa, line 50: This should not be hard-coded, but it should depend on the task executor.

The same holds for other setup lines (namely the cpu/memory requirements), which should be configurable from `fractal_server.app.runner` according to the content of `WorkflowTask.meta` or `Task.meta`.

fractal_server/config.py
```diff
@@ -152,6 +152,7 @@ def DATABASE_SYNC_URL(self):
     RUNNER_ROOT_DIR: Optional[Path]
     FRACTAL_LOGGING_LEVEL: int = logging.WARNING
     FRACTAL_PUBLIC_TASK_SUBDIR: str = ".fractal"
+    SLURM_PYTHON_WORKER_INTERPRETER: Optional[str] = None
```
tcompa, line 155: Can we add a comment to describe this variable?

tests/conftest.py
```diff
@@ -21,5 +21,29 @@ async def testdata_path() -> Path:
...
+        return tmp
+
+    yield _tmp_path_factory
+
+
+@pytest.fixture
+def tmp777_path(tmp_path):
+    tmp_path.chmod(0o777)
+    for parent in tmp_path.parents:
+        if "pytest" in parent.as_posix():
```
tcompa, line 42: I think this breaks if we use `pytest --basetemp $HOME/tmp`.
But it would work with `pytest --basetemp $HOME/tmp-pytest`, so I think there is no problem.

(nothing to do here)

tests/fixtures_server.py
```diff
@@ -54,13 +54,21 @@ def get_patched_settings(temp_path: Path):
 
     settings.FRACTAL_ROOT = temp_path
     settings.RUNNER_ROOT_DIR = temp_path / "artifacts"
+    settings.RUNNER_ROOT_DIR.mkdir(parents=True, exist_ok=True)
+    settings.RUNNER_ROOT_DIR.chmod(0o777)
+
+    # NOTE:
+    # This variable is set to work with the system interpreter within a docker
+    # container. If left unset it defaults to `sys.executor`
```
tcompa, line 62: typo executor->executable

tests/test_backend_slurm.py
```diff
@@ -0,0 +1,76 @@
...
+    from fractal_server.app.runner._slurm.executor import write_batch_script
+    import subprocess
+    import shlex
+
+    SBATCH_SCRIPT = "test"
+    f = write_batch_script(SBATCH_SCRIPT, script_dir=tmp777_path)
+
+    out = subprocess.run(
+        # f"sudo --non-interactive -u test01 cat {f}",
+        shlex.split(f"sudo --non-interactive -u test01 cat {f}"),
```
tcompa, line 54: This will fail on UZH, since `cat` is not part of the authorized sudo commands

fractal_server/app/runner/_common.py
```diff
@@ -165,3 +169,112 @@ def call_single_parallel_task(
...
+        # step 0: return future containing original task_pars
+        pseudo_future: Future = Future()
+        pseudo_future.set_result(task_pars)
+        return pseudo_future
+
+    logger = logging.getLogger(task_pars.logger_name)
+
+    # step n => step n+1
+    logger.debug(f"submitting task {this_task.order=}")
+    # parallelization_level = this_task.task.parallelization_level
```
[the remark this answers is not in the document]
jacopo-exact: yes

fractal_server/app/runner/_slurm/__init__.py
jacopo-exact: amended

fractal_server/app/runner/_slurm/executor.py
```diff
@@ -0,0 +1,228 @@
+# This adapts clusterfutures <https://github.com/sampsyo/clusterfutures>
+# Original Copyright
+# Copyright 2021 Adrian Sampson <asampson@cs.washington.edu>
+# License: MIT
+#
+# Copyright 2022 Jacopo Nepsolo <jacopo.nespolo@exact-lab.it>
```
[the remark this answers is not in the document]
jacopo-exact: Two issues:
1. this module is a derived work of Sampson's package, which is MIT licensed. We are thus required to keep the original license.
2. I'm starting to prefix the copyright with `#` wherever I see it because otherwise autodoc utilities tend to interpret it as module level documentation.

For these reasons, no change is needed.

fractal_server/app/runner/_slurm/executor.py
jacopo-exact: will solve #284 separately

fractal_server/app/runner/_slurm/executor.py
jacopo-exact: will solve #284 separately.

Also, it is not just here that might need changes. Also the sbatch script sets output and error for the job.

fractal_server/app/runner/_slurm/executor.py
jacopo-exact: it is intentional, as it is the original variable name from upstream.

fractal_server/app/runner/_slurm/executor.py
jacopo-exact: uncommented, tests passing locally.

fractal_server/app/runner/_slurm/executor.py
jacopo-exact: Let's discuss this. I'm not sure I understood the question.

fractal_server/app/runner/_slurm/__init__.py
jacopo-exact: deferred to #278

tests/conftest.py
jacopo-exact: Will now loudly fail if basetemp does not contain `pytest` in its name.

fractal_server/config.py
jacopo-exact: done

tests/fixtures_server.py
jacopo-exact: done

tests/test_backend_slurm.py
jacopo-exact: made more generic

fractal_server/app/runner/_slurm/executor.py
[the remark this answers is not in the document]
tcompa: My point was more that copyright is with FMI/UZH, and we list our names as authors.

> this module is a derived work of Sampson's package, which is MIT licensed. We are thus required to keep the original license.

Sure thing.
BTW, is it enough to mention it here? I guess a note in the main README is in order (copyright is with FMI/UZH, and module X is "Copyright 2021 Adrian Sampson")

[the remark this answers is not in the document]
jacopo-exact: I think as far as the license goes, it is enough that the README state "copyright by us unless otherwise stated in each individual module". Nevertheless, being nice is a free investment with high returns, so I shall add a comment in the readme.

fractal_server/app/runner/_slurm/executor.py
tcompa: As of `sbatch` man page, the `--parsable` flag can provide two kinds of output, either `jobid` or `jobid; clustername` (if there is a cluster name defined). If #169 moves forward, we may have cases where the cluster name _is_ defined, and then the output will be of the second type. Current code will break, and it's all right, but let's at least include a check that `output.stdout` can be cast to an integer. If not, we can just raise some error - and we'll fix it later when there is a relevant use case.

Note: this is already done upon return (where `jobid` is cast to `int`), but it would be nice to catch the error ourselves.

fractal_server/app/runner/_slurm/executor.py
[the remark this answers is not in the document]
tcompa: > Nevertheless, being nice is a free investment with high returns

Agreed. I'll also drop them an email, for information, in the next days

tcompa (approved):

tests/conftest.py
```diff
@@ -8,6 +8,19 @@
...
+def check_basetemp(tpath: Path):
+    """
+    Check that root temporary directory contains `pytest` in its name.
+
+    This is necessary because some tests use the directory name as a
+    discriminant to set and test permissions.
+    """
+    if "pytest" not in tpath.as_posix():
+        raise ValueError(
+            "`basetemp` must contain `pytest` in its name. Got {tpath.parent}"
```
tcompa, line 20: Missing `f` for f-string

PR #300, opened 2022-11-24
278 set clusterfutures slurm options

jacopo-exact:
- [x] serial tasks
- [x] parallel tasks

### DIFF
--- a/fractal_server/app/runner/_slurm/__init__.py
+++ b/fractal_server/app/runner/_slurm/__init__.py
@@ -9,6 +9,10 @@ from .._common import recursive_task_submission
 from ..common import async_wrap
 from ..common import TaskParameters
 from .executor import FractalSlurmExecutor
+from pydantic import BaseModel
+from ....syringe import Inject
+from ....config import get_settings
+import json
 
 
 """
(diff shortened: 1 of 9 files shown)


fractal_server/app/runner/_slurm/__init__.py
```diff
@@ -19,6 +25,75 @@
 """
 
 
+class SlurmConfig(BaseModel):
+    """
+    `name` maps on `Task.executor`.
```
tcompa, line 30: `name` is not in the list of class attributes. Is this on purpose?

tcompa: Ok, now I see that `SlurmConfig.name` was moved to the keys of `config_dict` (below).

Then we only need to update the docstring.

tcompa: BTW, we always want to use `WorkflowTask.executor`, and not `Task.executor` (it's only wrong in the docstring)

fractal_server/app/runner/_slurm/executor.py
```diff
@@ -154,6 +154,55 @@ def __init__(
         self.username = username
         self.script_dir = script_dir
 
+    def map(
+        self,
+        fn: Callable[..., Any],
+        *iterables,
+        timeout: Optional[float] = None,
+        chunksize: int = 1,
```
tcompa, line 162: What is `chunksize`?

fractal_server/app/runner/_slurm/executor.py
```diff
@@ -154,6 +154,55 @@ def __init__(
...
+        additional_setup_lines: Optional[List[str]] = None,
+    ):
+        """
+        Returns an iterator equivalent to map(fn, iter), passing
+        parameters to submit
+
+        This function overrides PSL's
+        `concurrent.futures.Executor.map` so that extra parameters
+        can be passed to `Executor.submi`.
```
tcompa, line 171: Typo in docstring (submi -> submit)

PR #322, opened 2022-11-30
delay backend selection

jacopo-exact:
this makes it easier for testing, but might make configuration error discovery slower for deployment as the system would only fail at the first use of the backend instead of at startup

### DIFF
--- a/tests/test_full_workflow.py
+++ b/tests/test_full_workflow.py
@@ -45,10 +45,6 @@ async def test_full_workflow(
     # Override RUNNER_BACKEND variable
     settings = get_patched_settings(tmp777_path)
     settings.RUNNER_BACKEND = backend
-    if backend == "slurm":
-        settings.FRACTAL_SLURM_CONFIG_FILE = (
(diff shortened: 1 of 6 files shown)


jacopo-exact:
@tcompa now it fails with a permission error on the sbatch file

tests/test_full_workflow.py
```diff
@@ -34,32 +31,25 @@ async def test_full_workflow(
...
+        RUNNER_BACKEND=backend,
+        FRACTAL_SLURM_CONFIG_FILE=testdata_path / "slurm_config.json",
+    )
 
     debug(f"Testing with {backend=}")
     if backend == "slurm":
         request.getfixturevalue("monkey_slurm")
         request.getfixturevalue("relink_python_interpreter")
-        request.getfixturevalue("slurm_config")
+        # request.getfixturevalue("slurm_config")
```
jacopo-exact, line 52: remove if not needed

jacopo-exact (commented):
A single very minor change if you really want

PR #327, opened 2022-12-01
106 handle task errors

### LINKED ISSUE #281: Runner (or backend) should expose `query_workflow_status`
- [ ] Each backend should update the status of a task (probably through `Executor` methods)
- [ ] The `query_workflow_status` function collects the status of all tasks

jacopo-exact:
- [x] STUB for monitoring api
- [x] basic error handling 

Close #281


fractal_server/app/api/v1/job.py
```diff
@@ -0,0 +1,26 @@
...
+router = APIRouter()
+
+
+@router.get("/{job_id}", response_model=ApplyWorkflowRead)
+async def get_job(
+    job_id: int,
+    user: User = Depends(current_active_user),
+    db: AsyncSession = Depends(get_db),
+) -> Optional[ApplyWorkflow]:
+    # FIXME acl
```
tcompa, line 23: Is access control something you'd like to include here? If not, let's open a new issue

jacopo-exact: will do it right now.

fractal_server/app/models/run.py
```diff
@@ -17,10 +17,11 @@
...
-class StatusType(str, Enum):
+class JobStatusType(str, Enum):
     SUBMITTED = "submitted"
-    PENDING = "pending"
+    RUNNING = "running"
     DONE = "done"
+    FAILED = "failed"
 
 
 class ApplyWorkflow(ApplyWorkflowBase, table=True):
```
tcompa, line 27: A docstring for `ApplyWorkflow` would be valuable

fractal_server/app/runner/__init__.py
```diff
@@ -55,14 +57,13 @@ def get_process_workflow():
...
     workflow: Workflow,
     input_dataset: Dataset,
     output_dataset: Dataset,
     job_id: int,
     username: Optional[str] = None,
     worker_init: Optional[str] = None,
-):
+) -> None:
     """
     Prepares a workflow and applies it to a dataset
```
tcompa, line 69: Docstring to be fixed (still lists `db` argument).

fractal_server/app/runner/__init__.py
```diff
@@ -77,6 +78,10 @@ async def submit_workflow(
         exist, a new dataset with that name is created and within it a new
         resource with the same name.
     """
+    db_sync = next(DB.get_sync_db())
+    job: ApplyWorkflow = db_sync.get(ApplyWorkflow, job_id)  # type: ignore
+    if not job:
+        raise ValueError("Cannot fetch job")
 
     input_paths = input_dataset.paths
     output_path = output_dataset.paths[0]
```
tcompa, line 87: Not related to current PR, but I am confused about the presence of `WORKFLOW_DIR` (here) and
```python

    @property
    def job_root_path(self) -> Path:
        settings = Inject(get_settings)
        return settings.RUNNER_ROOT_DIR / f"job_{self.id:06d}"
```
(in `app/models/run.py`).

fractal_server/app/runner/__init__.py
```diff
@@ -77,6 +78,10 @@ async def submit_workflow(
         exist, a new dataset with that name is created and within it a new
         resource with the same name.
     """
+    db_sync = next(DB.get_sync_db())
```
tcompa, line 81: For the record, what was the problem with the previous version (having db as an `AsyncSession` argument)?

fractal_server/app/runner/__init__.py
```diff
@@ -110,19 +115,36 @@ async def submit_workflow(
...
+            output_path=output_path,
+            input_metadata=input_dataset.meta,
+            username=username,
+            workflow_dir=WORKFLOW_DIR,
+            logger_name=logger_name,
+            worker_init=worker_init,
+        )
+
+        logger.info(f'END workflow "{workflow.name}"')
+        close_job_logger(logger)
```
tcompa, line 135: Logger should be closed also in case of failed execution.
In that case, we can also replicate the `job.log` contents in the logger.

tests/data/slurm_docker_images/node/Dockerfile
```diff
@@ -1,7 +1,9 @@
 FROM rancavil/slurm-node:19.05.5-1
 
-RUN pip3 install clusterfutures==0.4
-RUN pip3 install https://github.com/fractal-analytics-platform/fractal-server/releases/download/1.0.0b1/fractal_server-1.0.0b1-py3-none-any.whl
 COPY slurm.conf /etc/slurm-llnl/
+COPY requirements.txt .
+RUN pip3 install -r requirements.txt
```
tcompa, line 5: Could we avoid an harcoded `requirements.txt` file (since that's not what we use to install this package)?

If I first extract the tar.gz file to a folder, and then use
```
pip install .[slurm]
```
it looks like pip takes care of collecting all requirements - even without the `requirements.txt` file.

jacopo-exact: indeed it does, but it doesn't cache the building stage. By installing the requirements before the `pip install .` the whole layer is cached.

I do agree it is not elegant to have it hardcooded, and in fact it was a quick fix. Will open a whole issue dedicated to CI/docker improvements.

fractal_server/app/runner/__init__.py
jacopo-exact: deferred to #334

fractal_server/app/runner/__init__.py
jacopo-exact: modifying the job object sparked a conflict with an open session in the endpoint and some subquery that is not compatible with async sessions. 

Now I make sure that all endpoint DB session be closed before registering the background task. This should guarantee that relevant objects are only present in one session at a time.

fractal_server/app/runner/__init__.py
```diff
@@ -110,19 +124,37 @@ async def submit_workflow(
...
-    await db.commit()
+    except TaskExecutionError as e:
+        job.status = JobStatusType.FAILED
+        job.log = (
+            f"TASK ERROR:"
+            f"Task id: {e.workflow_task_id} ({e.task_name}), "
+            f"{e.workflow_task_order=}\n"
+            f"TRACEBACK:\n{str(e)}"
+        )
+        db_sync.merge(job)
```
tcompa, line 158: Repeating a previous comment: I think we should call `close_job_logger` also in this `except` branch (i.e. if some task failed).

PR #347, opened 2022-12-14
Add gunicorn dependency/example, and add example for fractal as a service

tcompa:


tcompa:
- [ ] Have gunicorn as an extra in pypoetry

Issue #366, opened 2022-12-20
Raising errors from a different thread (e.g. `executor.FileWaitThread`)


tcompa:
1. Case 1: a SLURM job fails for task-unrelated reasons (typical examples: out-of-memory kill by SLURM, or timeout, or even a `scancel`). As of https://github.com/fractal-analytics-platform/fractal-server/issues/343, we will capture and re-raise this error.
2. Case 2 (somewhat artificial, just for testing): the cloudpickle output file is created but it's corrupted, or its content is invalid for whatever reason.

These errors take place in the `FileWaitThread` thread attribute of the `FractalSlurmExecutor` (inheriting from the SlurmeExectutor of cfut, see [here](https://github.com/sampsyo/clusterfutures/blob/master/cfut/__init__.py)), either directly within `FileWaitThread.run()` or in its `callback` function.

How do we propagate those errors to the main thread, so that they can be transformed in a `TaskExecutionError` (or similar)?

---

Here is a minimal non-working example
```python
import threading
import time
import logging


class worker_thread(threading.Thread):
    def run(self):
        for i in range(2):
            logging.warning(f"Logging from worker_thread, {i}-th iteration")
            time.sleep(1)
        logging.warning("worker_thread will now raise")
        raise RuntimeError("this is an error")


class main_thread():
    def __init__(self):
        t = worker_thread()
        t.start()
        time.sleep(0.5)   # this is just to get nicer output, with separated main/worker logs
        for i in range(5):
            print()
            logging.warning(f"Logging from main_thread, {i}-th iteration")
            logging.warning(f"  Is worker_thread alive? {bool(t.is_alive())}")
            time.sleep(1)

main_thread()
```
with output
```
WARNING:root:Logging from worker_thread, 0-th iteration

WARNING:root:Logging from main_thread, 0-th iteration
WARNING:root:  Is worker_thread alive? True
WARNING:root:Logging from worker_thread, 1-th iteration

WARNING:root:Logging from main_thread, 1-th iteration
WARNING:root:  Is worker_thread alive? True
WARNING:root:worker_thread will now raise
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/home/tommaso/Fractal/test_threads/mnwe.py", line 12, in run
    raise RuntimeError("this is an error")
RuntimeError: this is an error

[...]

Issue #343, opened 2022-12-14
Handle SLURM errors


jluethi:
I ran a fractal workflow doing cellpose on a big 3D dataset and it ran out of memory on the cpu-mid executor. The following was in the slurm.err file:
```
slurm error contain:
srun: error: vcl1016: task 0: Out Of Memory
slurmstepd: error: Detected 1 oom-kill event(s) in StepId=130.0. Some of your processes may have been killed by the cgroup out-of-memory handler.
```

This also showed up on the server (in the console running the server), so I guess this is a state we're not handling well yet:
```
Exception in thread Thread-434:
Traceback (most recent call last):
  File "/ENV/FOLDER/fractal-server-v1-b4/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/ENV/FOLDER/fractal-server-v1-b4/lib/python3.8/site-packages/cfut/__init__.py", line 61, in run
    self.callback(self.waiting[filename])
  File "/ENV/FOLDER/fractal-server-v1-b4/lib/python3.8/site-packages/fractal_server/app/runner/_slurm/executor.py", line 338, in _completion
    fut.set_exception(exc)
  File "/ENV/FOLDER/fractal-server-v1-b4/lib/python3.8/concurrent/futures/_base.py", line 547, in set_exception
    raise InvalidStateError('{}: {!r}'.format(self._state, self))
concurrent.futures._base.InvalidStateError: CANCELLED: <Future at 0x7f778ce452e0 state=cancelled>
```

tcompa:
What fractal-server currently supports are tasks error, which are then encapsulated in `TaskExecutionError`s. This error is specific of the SLURM backend, but we can have a look and try to be more flexible.

---

Side-comment: good to know that FMI enforces memory hard limits through cgroup. This is not always the case (e.g. it's not like this at UZH), so that the error may be different depending on the SLURM config. My guess is that this would fail "correctly" at UZH, with an out-of-memory error from the task that is captured as a `TaskExecutionError`.

tcompa:
Reminder to self:
This can be tested by having a task that sleeps 10 seconds and setting a walltime limit of 1 second - so that the failure should be a SLURM one.

tcompa:
Another (non-equivalent) option: submit a job to an invalid partition.

tcompa:
For the record,

[...]

PR #367, opened 2022-12-20
[OLD] Handle slurm errors

tcompa:
To do (at the end):
- [ ] Put back the clusterfutures dependency as an extra

Ref:
* #366
* #343


tcompa:
If we manage to have the "standard" test suite (AKA all the tests which were already there before this PR) passing, then using this new version of clusterfutures should go very much in the intended direction of raising an error if something went wrong in the SLURM job. In that case we would just have to fix some details here and there.

PR #373, opened 2022-12-21
add download logs endpoint

### LINKED ISSUE #369: New monitoring endpoint: provide all logs
https://github.com/fractal-analytics-platform/fractal-server/issues/52#issuecomment-1359409604 covers the essential logging & user feedback by giving a way to see the error messages when something failed. A future enhancement would be to give users access to the logs for each task, e.g. to see if there were any warnings for a given task. I think the core thing here would be access to the .err files.

This is not urgent though. Will also come up as a discussion around the web frontend as well :)

mfranzon:
Close #369

### DIFF
--- a/fractal_server/app/api/v1/job.py
+++ b/fractal_server/app/api/v1/job.py
@@ -77,7 +77,7 @@ async def download_job_logs(
     byte_stream = BytesIO()
     with ZipFile(byte_stream, mode="w", compression=ZIP_DEFLATED) as zipfile:
         for fpath in working_dir_path.glob("*"):
-            zipfile.write(fpath)
+            zipfile.write(filename=str(fpath), arcname=str(fpath.name))
 
     return StreamingResponse(
         iter([byte_stream.getvalue()]),
--- a/fractal_server/app/api/v1/job.py
+++ b/fractal_server/app/api/v1/job.py
@@ -6,6 +6,7 @@ from fastapi import APIRouter
 from fastapi import Depends
 from fastapi import HTTPException
 from fastapi import status
+from fastapi.responses import StreamingResponse
 
 from ...db import AsyncSession
 from ...db import get_db
@@ -29,7 +30,7 @@ async def get_job(
     job = await db.get(ApplyWorkflow, job_id)
     if not job:
         raise HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND, detail="Project not found"
+            status_code=status.HTTP_404_NOT_FOUND, detail="Job not found"
         )
     await get_project_check_owner(
         project_id=job.project_id, user_id=user.id, db=db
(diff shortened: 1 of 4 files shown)


fractal_server/app/api/v1/job.py
```diff
@@ -46,3 +47,46 @@ async def get_job(
...
 
     return job_read
+
+
+@router.get("/download/{job_id}", response_class=StreamingResponse)
+async def download_job_logs(
+    job_id: int,
+    user: User = Depends(current_active_user),
+    db: AsyncSession = Depends(get_db),
+) -> Optional[ApplyWorkflow]:
```
tcompa, line 57: typo

fractal_server/app/api/v1/job.py
```diff
@@ -46,3 +47,46 @@ async def get_job(
...
+        project_id=job.project_id, user_id=user.id, db=db
+    )
+
+    job_read = ApplyWorkflowRead(**job.dict())
+
+    from devtools import debug
+
+    debug(job_read)
+
+    from io import BytesIO
```
tcompa, line 73: Move imports

fractal_server/app/api/v1/job.py
```diff
@@ -46,3 +47,46 @@ async def get_job(
...
+    from zipfile import ZipFile
+    from zipfile import ZIP_DEFLATED
+    import os
+
+    io = BytesIO()
+    zip_sub_dir = (
+        f"{job_read.project_id}_{job_read.workflow_id}_{job_id}_archive"
```
tcompa, line 80: Use `job_read.working_dir.name`

fractal_server/app/api/v1/job.py
```diff
@@ -46,3 +47,46 @@ async def get_job(
...
+    )
+    zip_filename = "%s.zip" % zip_sub_dir
+    with ZipFile(io, mode="w", compression=ZIP_DEFLATED) as zip:
+        for fpath in os.listdir(job_read.working_dir):
```
tcompa, line 84: `job_read.working_dir.glob()`

fractal_server/app/api/v1/job.py
```diff
@@ -46,3 +47,46 @@ async def get_job(
...
+            zip.write(job_read.working_dir / fpath)
+        # close zip
+        zip.close()
```
tcompa, line 87: Already taken care of by `with Zipfile() as zip`?

tcompa: I take it back.
But then let's use an instance rather than `with`

mfranzon: > ZipFile is also a context manager and therefore supports the [with](https://docs.python.org/3/reference/compound_stmts.html#with) statement. ...closed after the with statement’s suite is finished—even if an exception occurs...

my bad, it's already a context manager. Just remove .close() method

fractal_server/app/api/v1/job.py
mfranzon: I'm using `job.dict()["working_dir"] `so `.name` it is not available for strings.

PR #402, opened 2022-12-29
Remove dataset router and add/verify delete endpoints

### LINKED ISSUE #386: {project,workflow,dataset} delete endpoints are not complete
Current versions only remove single objects from the db, and not all the "children" objects.

mfranzon:
closes #386


tests/test_full_workflow.py
```diff
@@ -181,7 +181,7 @@ async def test_full_workflow(
 
         # Verify output
         res = await client.get(
-            f"{PREFIX}/dataset/{project_id}/{output_dataset_id}"
+            f"{PREFIX}/project/{project_id}/{output_dataset_id}"
```
tcompa, line 184: Note: this must be changed in the client as well: https://github.com/fractal-analytics-platform/fractal/issues/388

tcompa (approved):

tcompa:
Note: I changed the order of endpoints in `api/v1/project.py`, so that now they are sorted based on their address. This messes up the PR diff (sorry!), but it brings a lot of clarity when just glancing through the file. Let's do it also on other api files (in a different PR).

PR #405, opened 2022-12-30
Assert that server and worker have same {fractal-server,cfut} version

tcompa:


tcompa:
- [ ] Add a test

tcompa:
Closing since it's stale, let's restart later if needed.

PR #415, opened 2023-01-02
Restricted unpickling

tcompa:


tcompa:
Closing since it's stale, let's restart later if needed.

Issue #400, opened 2022-12-29
Should `add_task_to_workflow` return a `WorkflowTaskRead`?


tcompa:
Currently it returns the whole `WorkflowRead` object

PR #421, opened 2023-01-03
Make add_task return a WorkflowTaskRead (ref #400)

tcompa:


tcompa:
Close in favor of #428

Issue #447, opened 2023-01-13
Test `nfs4_setfacl`


tcompa:
As part of https://github.com/fractal-analytics-platform/fractal-server/issues/317#issuecomment-1352836505, we would need to devise a similar example for how to work with NFS4 ACLs at UZH.
Basic refs:
* https://linux.die.net/man/1/nfs4_setfacl
* https://www.osc.edu/book/export/html/4523

At the moment we have this script that works:
```
FOLDER=/some/folder/we/care/about
OTHERUSERID=something
CURRENTUSERID=something-else

rm -rf $FOLDER
mkdir $FOLDER
chmod 770 $FOLDER
echo

ls -la | grep mydir
echo

nfs4_setfacl -a A:d:$OTHERUSERID:RWX $FOLDER
nfs4_setfacl -a A:d:$CURRENTUSERID:RWX $FOLDER
echo

nfs4_getfacl $FOLDER
echo

ls -la | grep mydir
echo
```

and produces output like
```
drwxrwx---  2 fractal fractal      4096 Jan 13 08:21 mydir

A::OWNER@:rwaDxtTcCy
A::something-else:rwaDxtcy
A::something:rwaDxtcy
A::GROUP@:rwaDxtcy
A::EVERYONE@:tcy
D:fdi:OWNER@:rwaDx
A:fdi:OWNER@:tTcCy
A:fdi:something-else:rwaDxtcy
A:fdi:something:rwaDxtcy
A:fdi:GROUP@:tcy
A:fdi:EVERYONE@:tcy

drwxrwx---  2 fractal fractal      4096 Jan 13 08:21 mydir
```

At this point, the other user (i.e., not the `fractal` one) can do something like
```
echo "this is a log" > /some/folder/we/care/about/log.txt
```
and it works.
Now the `fractal` user can read/edit/remove the `log.txt` file, as expected.

Open issues:
- [ ] At the moment we need to work with the user IDs, which is not ideal, but that's probably due to some NFS configuration that we don't understand. If we can't solve it, we'll go with the shell command `id -u $USERNAME` and just extract it from there.
- [x] In principle a third user cannot do anything in the `/some/folder/we/care/about/` folder. Let's check it by hand.

tcompa:
Closing by now, in favor of #473.

Issue #317, opened 2022-11-29
Making slurm logs belong to the fractal user


jluethi:
Jumping on the work here: https://github.com/fractal-analytics-platform/fractal-server/issues/315 and extending on the discussion here: https://github.com/fractal-analytics-platform/fractal-server/issues/292

Is it possible to make the log files being owned by the fractal user?

This came up as part of the discussion around the deployment of Fractal at FMI today. Given that we're now more in control of those files with our own backend, having these mixed permissions can lead to some issues (i.e. the fractal user not being able to delete them).

Could we make those files be owned by the fractal user to avoid this?

We would still need to give the users access to write in them while the workflow is running (but could actually change the permissions when the workflow finishes to avoid users modifying or deleting log files). Enrico suggested the fractal user could create empty log files (touch filename.err) and give the users permission to write to them.
[the removing of write access later seems like an extra feature to me]

jluethi:
Notes from today's discussion:
Doing this directly will be tricky (not tested extensively). To avoid this being a blocking issue and problems removing logs as fractal user, we can set the permissions on the log files broadly for the time being.
One workaround if we need to achieve ownership by fractal user will be copying over the logs to a new, fractal-owned folder after the processing has finished. Adds extra complexity though, so let's see if it's needed.

jluethi:
And another idea:
Using `setfacl` to give the specific user that's running a job access to just that folder
i.e.:

> Fractal would need to set the correct ACL for each job folder for the correct user. At least this would be the best solution. If easier you can bodge it by moving the artifacts (and any other folder the user need to write into) to /tmp/<fractal-subfolder> and then copy them over once the job is done. But honestly the first option sounds much simpler as you need to add pretty much a single line of code to fractal, something like subprocess.run(['setfacl', blablabla])

tcompa:
We had a look at the `setfacl` plan, with @mfranzon.

TL;DR
* It seems to "work" for our intended use (note: ownerships are not modified, but access is much more strict).
* Its vanilla version does not work on the nfs4 UZH filesystem, because it would require us to use `nfs4_setfacl` command.

---

### Preliminary: which filesystem?

* We work in the `/tmp` folder.
* `df /tmp` shows that it is part of `/dev/vda1`
* `mount | grep "/dev/vda1"` shows that this is an `ext4` filesystem.
* On the contrary, using `/data/homes/fractal` as a folder points to a `nfs4` filesystem. We discard that case by now.

[...]

PR #449, opened 2023-01-13
Implement access control list for job folders

### LINKED ISSUE #448: Implement Access Control List for job folders
## Requirements

Every time a workflow is applied, a folder is created for the corresponding job (a typical name would look like `workflow_0007_job_0002`). 

In a single-user ("user" as in "user of the machine") use case, there is nothing special about this folder. In a SLURM multi-user case, there are two different users ("users" as in "users of the cluster") that need to have read/write access to this folder, as part of this scheme:
1. `fractal`  (or whatever user is running the server) creates the folder
2. `fractal` writes some logs.
3. `fractal` writes some files that are necessary for running the task (e.g. the sbatch script and a serialized version of the task to run).
5. `test01` (or whatever user needs to run the task) reads the files from point 2, and runs the task.
6. `test01` writes some logs, and some files with the task output.
7. `fractal` reads the output files and proceeds.
8. `fractal` need to be able to access any other log files (created by `test01`) in that folder.

## Current status
At the moment, `fractal` provides broad (777) access to the job folder, and `test01` provides broad (777) access to their logs. This works, but it's obviously not the way to go (ref #317)

## Proposed change

We add a new config variable (provisional name `FRACTAL_ACL_OPTIONS` defaulting to `none`), and proceed differently based on its value:
* If `FRACTAL_ACL_OPTIONS="none"`, folders and files are created with standard permissions (i.e. we never set them to 777). This is meant to only work for a single-user use case, where the user has access to all of their folders/file.
* If `FRACTAL_ACL_OPTIONS="standard"` (or a better name..), we assume that the `setfacl` command is available on the machine running the server (we can also check it by hand, to be sure), and we proceed as in https://github.com/fractal-analytics-platform/fractal-server/issues/317#issuecomment-1352836505. That is, we use standard ACLs to provide access to the job folder to **two users only** (`fractal` and `test01`). This is done upon folder creation, and then we don't touch permissions any more (neither on the `fractal` side, nor on the `test01` side).
* If `FRACTAL_ACL_OPTIONS="nfs4"`, we follow the same logic but with nfs4-specific commands (as in #447).

tcompa:
Close #448
Ref #317
Ref #447


tcompa:
Closing this PR, since #448 was replaced by #474.

PR #450, opened 2023-01-13
Revamp users management

### LINKED ISSUE #403: Restrict user registration to superuser
Some actions must be only possible for a fractal superuser (any user can be created as a superuser). This issue concerns access control to this kind of actions (a typical example would be removing tasks from the db - branching from #386).

### LINKED ISSUE #442: Add tests for `fastapi_users`'s `users` routers
From https://fastapi-users.github.io/fastapi-users/10.2/usage/routes, we should test

```
# Super user only:
POST /register
GET /{user_id}
PATCH /{user_id}
DELETE /{user_id}

# All users:
GET /me (already tested)
```

This should be part of #441

### LINKED ISSUE #446: Revamp users management
It's probably betterto have a fresh issue for this feature, related to:
* #403
* #442
* #129 

In our current plan, the way users-management works would be the following:
1. At server startup, a Fractal user ("Fractal user" as in "user of the Fractal server") is created in the database, with default credentials (e.g. username `admin` and password `fractal-is-great`). `admin` is a Fractal superuser. Note that this user doesn't have any `slurm_user` attribute set.
2. Depending on your use-case, different paths are taken:
    * If you want to use Fractal on your own (that is, on your machine, and with no multiple users), you won't need to register any other user. You'll write the default credentials in your client environment variables (or `.env` file), and you are ready to start.
    * If you want to use Fractal in a multi-user way and/or on SLURM, more steps are needed. (A) Acting via the client (`fractal user edit` command) you need to edit the `admin` user (at least to change its password); (B) acting via the client (`fractal user register`), you need to register any other user that should access Fractal. Note that 

The first use case does not require many comments.
Concerning the multi-user use  case:
* Note that this also solves #129, for the moment, because the association between Fractal and SLURM usernames is guaranteed by the admin.
* Registration of new users won't any more be open (as it would be in a public service), but it will be only allowed to the a superuser - ref #403. The same is true for most of the `/users/` endpoints, with the only exception of `/users/me` (which gives to any user the possibility to see their details).
* There can be as many superusers as necessary. One is always the `admin` user, created on startup, but others can exist.

---

Open questions (@mfranzon): what happens on re-starts of the server? Let's make sure we don't try to recreate the `admin` user if it already exists.

mfranzon:
Close #446
Close #403
Close #442


tcompa:
Reminder:
- [ ] Make slurm_user an optional field of the User model

tcompa:
Is this first user an actual superuser?

I'm still trying to test it, but see https://fastapi-users.github.io/fastapi-users/10.2/usage/flow/?h=superuser#5-become-a-superuser:
> The very first superuser can only be set at database level: open it through a CLI or a GUI, find your user and set the is_superuser column/property to true.

tcompa:
Since the "startup" logic is not executed in the tests, I added a `registered_superuser_client` fixture which is supposed to produce a superuser. The current version fails.

tcompa:
> Since the "startup" logic is not executed in the tests, I added a `registered_superuser_client` fixture which is supposed to produce a superuser. The current version fails.

Fixed this by actually using the new `_create_user` function in the fixture.

fractal_server/main.py
```diff
@@ -61,6 +75,44 @@ async def __on_startup() -> None:
...
 
 
+async def _create_user(
+    email: str,
+    password: str,
+    is_superuser: bool = False,
+    slurm_user: Optional[str] = None,
+) -> None:
+    """
+    Private method to create a fractal user
```
tcompa, line 85: Add link to fastapi_users docs

Issue #473, opened 2023-01-26
Separate server/user job-execution folders


tcompa:
(based on several ongoing discussions, and edited accordingly)

---

# Files

Two kinds of files should be "exchanged" between server and cluster user

1. Files written by fractal (AKA the user running the server), from python:
    * JSON file with task input &#x1F534;
    * Sbatch scripts
    * Pickle file with cfut job input
    * Log files

2. Files written by test01
    * Log files for the tasks (written by test01, from python)
    * JSON file with task output (written by test01, from python)
    * SLURM stdout/stderr (written by SLURM)
    * Pickle file with cfut job output (written by test01, from python)

# sudo requirements:
* fractal can run `sudo -u test01 sbatch something.sh`
* fractal can run `sudo -u test01 cat something.sh`
* fractal can run `sudo -u test01 ls something.sh`

# Folder structure

### Current version
JOBDIR and USERJOBDIR are the same folder, owned by fractal and with 777 permissions. All the rest is the same as in future version.

### Future version:

Folders owned by fractal:
* ARTIFACTS=/some/share/fractal/folder/artifacts                          755, visible to all users
* JOBDIR=/some/share/fractal/folder/artifacts/workflow_01_job_02          755, visible to all users
* TASKDIR=/some/share/fractal/folder/tasks                                755, visible to all users

Folder owned by user (i.e. test01):
* PROJECT_DIR=/some/share/test01/folder                                   700, but visible to fractal through sudo-cat
* USERJOBDIR=/some/share/test01/folder/tmp_wf_01_job_01   700, but visible to fractal through sudo-cat

Notes:
* PROJECT_DIR must exist (check upon project creation)
* USERJOBDIR must exist (mkdir at the beginning of submit.sh) &#x1F534;

# Workflow execution

### Current version
* fractal writes args.json and submit.sh input.pickle to JOBDIR (and also logs, but that doesn't matter because test01 has no access to workflow.log)

### Future version
* same as current version

### Current version

Steps:
* fractal runs `sudo -u test01 sbatch JOBDIR/submit.sh`
  (OK, since JOBDIR is 777)
* test01 (via sbatch) runs `srun /some/path/python -m cfut JOBDIR/input.pickle`
  (OK, since JOBDIR is 777)
* test01 (via srun) runs `TASKDIR/some/path/python TASKDIR/some/path/task.py --input JOBDIR/args.json`
  (OK, since JOBDIR is 777)
[...]

PR #475, opened 2023-01-26
 Separate server/user job-execution folders

### LINKED ISSUE #470: Raise error if `slurm_user` is not set (when using the slurm backend and submitting a workflow)
Currently we only fail due to some "permission denied" during tasks, but we could catch the error much earlier.

tcompa:
Ref #473 
Close #470 (see e.g. 73b944c9bd56819c9eb1f91b98fcbe1b550dcfc4)

See https://github.com/fractal-analytics-platform/fractal-server/issues/473#issuecomment-1415941794 for missing FIXMEs.


mfranzon (approved):

tcompa:
There's likely some issue with github CI, since tests now take quite longer than the last ones on this same branch.

As a check that this is not a problem with the new code, here are timings for the whole CI (running locally) at two recent points in history:

* 8300e004 (current): 168 s
* 45518131 (tag 1.0.3a2): 161 s

This looks under control, independently on github CI runtimes.

PR #478, opened 2023-01-27
Remove srun from python-cfut command in sbatch script (close #477) and update poetry to 1.3.2

### LINKED ISSUE #477: Remove `srun` from SLURM submission script?
The submission script we use currently has a line like
```
srun /some/python -m fractal_server.app.runner._slurm.remote /some/path/cfut.in.xxx.pickle 
```

Some unexpected error appeared in the local CI (something related to runpy failing upon a certain import from sqlalchemy), which is unclear (since logging in into the container and importing that same module works). Removing `srun` from the sbatch script may actually solve this issue, since that one would be same command that works from the terminal from inside the container.

Besides, this raises a relevant question: is the `srun` part doing anything special there? I think it doesn't, since this is already part of a sbatch submission (and since we don't have to combine different commands, or to assign special slurm options). We can probably remove it, independently on the CI issue mentioned above.

(to be verified)

tcompa:


tcompa:
Since this turned out to not being the actual CI-failure reason, I am putting `srun` back as part of #475 (see bf83675ec32edbb2af6447bfc6045f9603f51c58).

This is likely not necessary, but it's just to avoid handling it at this point.

PR #490, opened 2023-02-06
Import/export workflows

### LINKED ISSUE #489: Pydantic `dict()` method for SQLModel objects does not include relationships
While working on #1 with @mfranzon , we hit an unexpected behavior of the [`.dict()` pydantic method](https://docs.pydantic.dev/usage/exporting_models
) when combined with `SQLModel` objects, where it only exports fields (e.g. `workflow.id`) and not relationships (e.g. `workflow.task_list`).

Because of the limited of scope of our needs (we "only" need to transform a Workflow object in a nested python dictionary and to remove the ID attributes, and possibly we'll do something similar for a Project or another model), we will now implement a simple custom solution for #1 on the line of
```python
def export(wf: Workflow) -> dict:
    wfread = WorkflowRead(**wf.__dict__)
    wfread_dict = wfread.dict(exclude={"id", "project_id"})
    for ind, wftask in enumerate(wfread_dict["task_list"]):
        wftask.pop("id")
        wftask.pop("task_id")
        wftask.pop("workflow_id")
        wftask["task"].pop("id")
        wfread_dict["task_list"][ind] = wftask
    return wfread_dict
```

Below are some relevant references, in case we want/need to improve this in the future:

One possible explanation is that the current behavior is needed to avoid infinite recursion, and the solution is to introduce additional (more explicit, and non-SQLModel-inherited) models. See
* https://sqlmodel.tiangolo.com/tutorial/fastapi/relationships/#models-with-relationships
* https://github.com/tiangolo/sqlmodel/pull/445 (this open PR also has useful minimal-failing-examples)

Other possibly related issues:
* https://github.com/tiangolo/sqlmodel/issues/444
* https://github.com/tiangolo/sqlmodel/issues/347
* https://github.com/tiangolo/sqlmodel/issues/63

A possibly related issue:
* https://github.com/fractal-analytics-platform/fractal-server/issues/488

tcompa:
Ref #1
Close #489


tcompa:
Updates to common:
* https://github.com/fractal-analytics-platform/fractal-common/commit/255df76de380234092ed0d08c017a430b1f2ef63
* https://github.com/fractal-analytics-platform/fractal-common/commit/5ac0d4ed01f11a61e60df5604ba9968c51a8d7bf

fractal_server/app/api/v1/workflow.py
```diff
@@ -125,6 +128,38 @@ async def _get_workflow_task_check_owner(
     return workflow_task, workflow
 
 
+async def _check_new_workflow_name_and_project_id(
```
mfranzon, line 131: change name from `_check_new_workflow_name_and_project_id` to `_check_workflow_exists`

mfranzon (approved):

PR #526, opened 2023-02-22
task_update endpoint partially fixed

### LINKED ISSUE #192: what does it mean to patch / update a task
as tasks are now treated as executables and automatically created from the manifest, it makes less sense to allow arbitrary changes to the Task db records.

A user may still want to change the name, but definitely should not change the command or the source.

This issue is just to remember that we'll have to think about this.

ychiucco:
closes #192

### DIFF
--- a/fractal_server/app/api/v1/task.py
+++ b/fractal_server/app/api/v1/task.py
@@ -302,7 +302,10 @@ async def patch_task(
 
     db_task = await db.get(Task, task_id)
 
-    for key, value in task_update.dict(exclude_unset=True).items():
+    update = {
+        k: v for k, v in task_update.dict().items() if v and (k != "source")
+    }
+    for key, value in update.items():
         if isinstance(value, str):
             setattr(db_task, key, value)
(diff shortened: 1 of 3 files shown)


tcompa:
Nice, thanks!
After the endpoint is complete, let's add a test (where we both edit e.g. the name **and** a dictionary field).

fractal_server/app/api/v1/task.py
```diff
@@ -293,23 +293,32 @@ async def patch_task(
...
+        #    current_default_args.update(value)
+        #    setattr(db_task, key, current_default_args)
         else:
             raise HTTPException(
                 status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-                detail="patch_task endpoint cannot set {key=}",
+                detail=f"patch_task endpoint cannot set {key=}",
             )
 
+    await db.merge(db_task)
```
tcompa, line 321: Is this needed? I think the endpoint was working even without it, but let's use the tests to check it.

Issue #498, opened 2023-02-09
Tests fail if python3.8 is not installed


tcompa:

tcompa:
The issue comes from this line in a fixture
```python
check_python_has_venv("python3.8", python_test_path)
```

@ychiucco

tcompa:
Ref #520

tcompa:
This should be fixed via https://github.com/fractal-analytics-platform/fractal-server/pull/527/commits/3a2e47a7dfc429d669ecd9504c4e7e9fe3f4c058

tcompa:
After some digging, it turns out that this is hard to solve due to the twofold use of our tests (testing tasks running locally or via slurm on a docker slurm cluster).
The `install_dummy_packages` fixture is where the problem appears explicitly:
* Setting `python_version=3.9` for tasks will lead to failures if python3.9 or its venv module are not installed.
* Setting it to `None` will lead to failures when running in the slurm docker, because it may point to (e.g.) python3.10 which is then not available in the slurm docker containers.

The goal (letting someone without python3.9 run the CI) is not worth the effort (splitting the fixture into two functions, one for each case), IMHO.

Won't fix.

PR #527, opened 2023-02-23
Drop python 3.8 support

tcompa:


tcompa:
The solution, for the CI and to avoid #498, was to always also install python 3.9 in the CI container - unless it's already there.

mfranzon (approved):

PR #567, opened 2023-03-15
Solves sqlite3.IntegrityError

### LINKED ISSUE #560: Fix `User`/`Project` models, due to unsupported UUID4 column in sqlite
After one project is deleted, trying to create a new project leads to an error for that user:
```
(sqlite3.IntegrityError) UNIQUE constraint failed: linkuserproject.project_id, linkuserproject.user_id\n[SQL: INSERT INTO linkuserproject (project_id, user_id) VALUES (?, ?)]\n[parameters: (6, '9cf73131431041a998ef494fdc99c230')]\n(Background on this error at: https://sqlalche.me/e/14/gkpj)
```

It still works for other users. But it appears deleting projects puts the user on that server into a compromised state.
Tested both on the UZH deployment and on a local server running on my Mac. But were running 1.1.0a1 version of the server.

If there is a good fix for it, let's fix this soon. Otherwise, let's deactivate the project delete functionality for the time being.

cc @adrtsc I could also reproduce your error here. We'll track progress on it in this issue. Thanks for finding it!

ychiucco:
closes #560


tcompa:
NOTE: the test passes on the github CI (with postgres) but it always fails locally (with sqlite) - ref #570

tcompa:
We now understand better the source of this error, see https://github.com/fractal-analytics-platform/fractal-server/issues/560#issuecomment-1472297432.
Closing this PR in favor of a broader refactor of the User models - ref #575.

PR #593, opened 2023-03-29
use `_get_dataset_check_owner`

### LINKED ISSUE #554: Use `_get_dataset_check_owner` in `patch_dataset` endpoint
This should replace the current block
```python
    project = await _get_project_check_owner(
        project_id=project_id, user_id=user.id, db=db
    )
    db_dataset = await db.get(Dataset, dataset_id)
    if db_dataset not in project.dataset_list:
        raise HTTPException(
            detail=f"Dataset {dataset_id} is not part of project {project_id}",
        )
```

ychiucco:
closes #554

### DIFF
--- a/fractal_server/app/api/v1/project.py
+++ b/fractal_server/app/api/v1/project.py
@@ -433,15 +433,11 @@ async def patch_dataset(
     """
     Edit a dataset associated to the current project
     """
-    project = await _get_project_check_owner(
-        project_id=project_id, user_id=user.id, db=db
+    db_dataset = await _get_dataset_check_owner(
+        project_id=project_id,
+        dataset_id=dataset_id,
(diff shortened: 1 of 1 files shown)


tcompa:
CI is broken, because `db` must be a required argument of the `_get_dataset_check_owner` function.

- [ ] Please remove the `Depends(get_db)` default; I think it is just wrong for this kind of function (as opposed to an endpoint function).
- [ ] Same thing for `_get_project_check_owner`.

tcompa:
Note that `_get_dataset_check_owner` returns a dictionary with both the dataset and the project, and not just the dataset.

- [ ] Please also update the type hint (`-> Dataset` --> `dict[str,Union[Dataset,Project]]`)

PR #594, opened 2023-03-30
Add test for non-python task

### LINKED ISSUE #189: Add test for non-python task
* Write a minimal bash script conforming to the the right interface (`./script.sh -j /path/to/args.json -o /path/to/out.json`)
* Construct the task object, and set the appropriate `command`.
* Run it through fractal-server, e.g. as in `test_runner_backends.py`

ychiucco:
closes #189


tests/test_task_api.py
```diff
@@ -371,3 +371,65 @@ async def test_task_collection_api_failure(
...
+            project_id=project.id,
+            input_dataset_id=input_dataset_id,
+            output_dataset_id=output_dataset_id,
+            workflow_id=workflow["id"],
+            overwrite_input=False,
+        )
+        debug(payload)
+        res = await client.post("/api/v1/project/apply/", json=payload)
+        debug(res.json())
+        assert res.status_code == 202
```
tcompa, line 435: We should add an assertion about the output. For instance we could modify the task so that it reads the input file (`--json`) and writes it to the output file (`--metadata-out`), and then check that the output file is equal to the input file.

tests/test_task_api.py
```diff
@@ -371,3 +371,65 @@ async def test_task_collection_api_failure(
         assert data["status"] == "fail"
         assert data["log"]  # This is because of verbose=True
         assert "fail" in data["log"]
+
+
+async def test_non_python_task(
```
tcompa, line 376: We are not really testing the task API here, but rather the workflow execution.
Let's move this test to a new file, named e.g. `test_full_workflow_bash_task.py`. Or we can also include it in `test_full_workflow.py`.

tcompa (changes requested):

fractal_server/app/api/v1/project.py
```diff
@@ -248,7 +247,10 @@ async def apply_workflow(
     else:
         try:
             output_dataset = await auto_output_dataset(
-                project=project, input_dataset=input_dataset, workflow=workflow
+                project=project,
```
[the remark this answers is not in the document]
ychiucco: Not at all. My fault, thank you

tests/test_full_workflow.py
```diff
@@ -443,3 +445,84 @@ async def test_failing_workflow_JobExecutionError(
...
+            input_dataset_id=input_dataset_id,
+            output_dataset_id=output_dataset_id,
+            workflow_id=workflow["id"],
+            overwrite_input=False,
+        )
+        debug(payload)
+        res = await client.post("/api/v1/project/apply/", json=payload)
+        debug(res.json())
+        assert res.status_code == 202
+
```
tcompa, line 521: You should add some waiting time here.
Simplest solution: `time.sleep(5)` (just for debugging)
Better solution: `while` loop which calls the `get_job` endpoint several times, until `status="done"`.

tcompa: Or, at least, something similar to:
```python
        res = await client.post("/api/v1/project/apply/", json=payload)
        job_data = res.json()
        debug(job_data)
        assert res.status_code == 202

        res = await client.get(f"{PREFIX}/job/{job_data['id']}")
        assert res.status_code == 200
        job_status_data = res.json()
        debug(job_status_data)
        assert job_status_data["status"] == "done"
```
so that the failure will be more informative.

tests/test_full_workflow.py
```diff
@@ -443,3 +445,91 @@ async def test_failing_workflow_JobExecutionError(
...
+        payload = {"name": "WF", "project_id": project.id}
+        res = await client.post("api/v1/workflow/", json=payload)
+        workflow = res.json()
+        debug(workflow)
+        assert res.status_code == 201
+
+        file_in = f"{str(testdata_path)}/test_in.json"
+        file_out = f"{str(testdata_path)}/test_out.json"
+
+        if os.path.exists(file_in):
```
tcompa, line 471: Using tmp_path will make this test not necessary.

tests/test_full_workflow.py
```diff
@@ -443,3 +445,91 @@ async def test_failing_workflow_JobExecutionError(
...
+            with open(file_in, "r") as f:
+                TEST_JSON = json.load(f)
+        else:
+            TEST_JSON = {"test": "json"}
+            with open(file_in, "w") as f:
+                json.dump(TEST_JSON, f)
+
+        task_dict = dict(
+            name="non-python",
+            source="custom task",
```
tcompa, line 481: Remove space

tests/test_full_workflow.py
```diff
@@ -443,3 +445,91 @@ async def test_failing_workflow_JobExecutionError(
...
+            command=(
+                f"sh {str(testdata_path)}/issue189.sh "
+                f"--json {file_in} --metadata-out {file_out}"
```
tcompa, line 484: Remove `--json` and `--metadata-out` from `command`, which will be added by fractal-server.

tcompa (approved):

tcompa:
The CI also fails locally, on my machine. I will debug it.

tcompa:
Current errror is due to a problem with the shell script:
```
/home/tommaso/Fractal/fractal-server/tests/data/non_python_task_issue189.sh: 11: [[: not found
/home/tommaso/Fractal/fractal-server/tests/data/non_python_task_issue189.sh: 32: [[: not found
```

tcompa:
Here is the reason: https://stackoverflow.com/a/3401197/19085332

PR #625, opened 2023-04-17
Expose local user's exports in SLURM configuration file

### LINKED ISSUE #605: Provide default WORKER_INITs in some server config
We have the flexible `WORKER_INIT` to handle setting environment variables for cache directories, which has proven to be very useful. We should keep this as something the user can add to.

But for our default workflows using fractal-tasks-core, the settings for `WORKER_INIT` are always the same. Typically, they are:
```
BASE_CACHE_DIR=${HOME}/.cache
WORKER_INIT="\
export CELLPOSE_LOCAL_MODELS_PATH=$BASE_CACHE_DIR/CELLPOSE_LOCAL_MODELS_PATH
export NUMBA_CACHE_DIR=$BASE_CACHE_DIR/NUMBA_CACHE_DIR
export NAPARI_CONFIG=$BASE_CACHE_DIR/napari_config.json
export XDG_CONFIG_HOME=$BASE_CACHE_DIR/XDG_CONFIG
export XDG_CACHE_HOME=$BASE_CACHE_DIR/XDG
"
```

And at FMI, we also specify some proxy settings to allow model download from cellpose.

----

It would make sense to abstract this away from the user, so that they don't need to set those variables themselves every time. Especially with a view towards submission from the web interface (see https://github.com/fractal-analytics-platform/fractal-web/issues/86), it doesn't make much sense to have users enter this in the web frontend.

Thus, we'd want to get to a point where we can have preset WORKER_INITs which are always run first, followed by optional additional WORKER_INITs provided by the user.

This comes with 3 questions:
1. Where do we configure those WORKER_INITs such that it can be set up for a given server deployment? 
2. How do we still make the cache-dir user-specific (because the user needs to be able to write there).
3. How do default WORKER_INITs interact with user-specified WORKER_INIT

1. could be some json file or something similar provided on server start-up
2. I think the answer would be using the newly created `user.cache_dir`. If we can set the `BASE_CACHE_DIR=${user.cache_dir}/.cache` and `user.cache_dir` is also provided as part of the WORKER_INIT, then that would be a great default!
3. I think we should maintain a user-defined WORKER_INIT. Typical use-cases: Someone hits a new edge-case where they need an env-variable set, e.g. during task development or by using tasks in a way we hadn't tested before. I think if we just add the user-provided `WORKER_INIT` after the server-default, then the user could even overwrite those cache folders if they wanted to do so

---

If we're able to get rid of the standard WORKER_INIT, that will be a great way of simplifying both the example scripts and making the web submission easier :)

tcompa:
Close #605

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Drop `project.project_dir` and replace it with `user.cache_dir` (\#601).
 * Update SLURM backend (\#582, \#612, \#614); this includes (1) combining several tasks in a single SLURM job, and (2) offering more granular sources for SLURM configuration options.
+* Expose local user exports in SLURM configuration file (\#625).
 * Review logging configuration (\#619, \#623).
 * Update to fastapi `0.95` (\#587).
 * Minor improvements in dataset-edit endpoint (\#593) and tests (\#589).
--- a/docs/internals/runners/slurm.md
+++ b/docs/internals/runners/slurm.md
@@ -34,6 +34,33 @@ cpus_per_task=3
 mem=10G
 ```
 
+### Exporing environment variables
+
+In a typical use case, the tasks are Python scripts that are executed by a
+certain user A, but with a Python interpreter that belongs to the user B who
+runs `fractal-server. For this reason, it may be needed to explicitly set some
+environment variables needed by external libraries, so that they are writeable
+by user A.
+
+The `fractal-server` admin can set some defaults for variables that must be
+exported. By including a block like
+```JSON
+{
+  ...
+  "user_local_exports": {
+    "var1": "path1",
+    "var2": "path2.json"
+  }
+```
+to the [SLURM configuration
+file](../../../reference/fractal_server/app/runner/_slurm/_slurm_config/#fractal_server.app.runner._slurm._slurm_config.SlurmConfigFile), the SLURM submission script will include the lines
+```bash
+...
+export var1=/this/is/the/cache/dir/of/user/A/path1
+export var2=/this/is/the/cache/dir/of/user/A/path2.json
+...
+```
+
 ## SLURM batching
 
 The SLURM backend in `fractal-server` may combine multiple tasks in the same
--- a/docs/internals/runners/slurm.md
+++ b/docs/internals/runners/slurm.md
@@ -51,6 +51,7 @@ exported. By including a block like
     "var1": "path1",
     "var2": "path2.json"
   }
+}
 ```
 to the [SLURM configuration
 file](../../../reference/fractal_server/app/runner/_slurm/_slurm_config/#fractal_server.app.runner._slurm._slurm_config.SlurmConfigFile), the SLURM submission script will include the lines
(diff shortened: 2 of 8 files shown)


tcompa:
The new docs would read as reported below. @jluethi does this sound like a reasonable solution?

---
### Exporing environment variables

In a typical use case, the tasks are Python scripts that are executed by a certain user A, but with a Python interpreter that belongs to the user B (that is, the `fractal-server` admin). For this reason, it may be needed to explicitly set some environment variables needed by external libraries, so that they are writable by user A.

The `fractal-server` admin can set some defaults for variables that must be exported. By including a block like
```JSON
{
  ...
  "user_local_exports": {
    "var1": "path1",
    "var2": "path2.json"
  }
```
to the [SLURM configuration file](../../../reference/fractal_server/app/runner/_slurm/_slurm_config/#fractal_server.app.runner._slurm._slurm_config.SlurmConfigFile), the SLURM submission script will include the lines
```bash
...
export var1=/this/is/the/cache/dir/of/user/A/path1
export var2=/this/is/the/cache/dir/of/user/A/path2.json
...
```

jluethi:
> Exporing environment variables

Should be Exporting, right?

I like the approach. We should make it more explicit that each variable gets the `cache_dir` prepended, thus this can only be used to export cache paths, not arbitrary environment variables.
And that the paths are relative to the cache directory.

tcompa:
Thanks. Fixed it as below, I think this is now ready.

---

### Exporting environment variables

The `fractal-server` admin may need to set some global variables that need to be included in all SLURM submission scripts; this can be achieved via the `extra_lines` field in the SLURM configuration file, for instance as in
```JSON
{
  "default_slurm_config": {
    "partition": "main",
    "extra_lines": [
      "export SOMEVARIABLE=123",
      "export ANOTHERVARIABLE=ABC"
    ]
  }
}
```

There exists another use case where the value of a variable depends on the user who runs a certain task. A relevant example is that user A (who will run the task via SLURM) needs to define the cache-directory paths for some libraries
they use (and those must be paths where user A can write).  This use case is also supported in the specs of `fractal-server` [SLURM configuration file](../../../reference/fractal_server/app/runner/_slurm/_slurm_config/#fractal_server.app.runner._slurm._slurm_config.SlurmConfigFile): If this file includes a block like
```JSON
{
  ...
  "user_local_exports": {
    "LIBRARY_1_CACHE_DIR": "somewhere/library_1",
    "LIBRARY_2_FILE": "somewhere/else/library_2.json"
  }
}
```
then the SLURM submission script will include the lines
```bash
...
export LIBRARY_1_CACHE_DIR=/my/cache/somewhere/library_1
export LIBRARY_2_FILE=/my/cache/somewhere/else/library_2.json
...
```
Note that all paths in the values of `user_local_exports` are interpreted as relative to a base directory which is user-specific (for instance `/my/cache/`, in the example above), and which is defined in the `User.cache_dir` attribute. Also note that in this case `fractal-server` only compiles the configuration options into lines of the SLURM submission script, without performing any check on the validity of the given paths.

jluethi:
Looks great, thanks for writing this extended version @tcompa !

PR #658, opened 2023-05-02
Use FRACTAL_SLURM_WORKER_PYTHON for venv/pip operations on tasks

tcompa:
- [ ] Depending on decisions in #659, keep/remove the check about simultaneous setting of version&FRACTAL_SLURM_PYTHON_WORKER.


tcompa:
This attempt has to be restarted afresh in view of #686

Issue #661, opened 2023-05-02
Task collection broken with sqlite?


tcompa:
Version `1.2.4` seems to introduce an issue with task collection, when using sqlite:
```python traceback
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/tommaso/miniconda3/envs/fractal-server-1.2.4/lib/python3.9/site-packages/uvicorn/protocols/http/h11_impl.py", line 407, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/tommaso/miniconda3/envs/fractal-server-1.2.4/lib/python3.9/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/home/tommaso/miniconda3/envs/fractal-server-1.2.4/lib/python3.9/site-packages/fastapi/applications.py", line 276, in __call__
    await super().__call__(scope, receive, send)
  File "/home/tommaso/miniconda3/envs/fractal-server-1.2.4/lib/python3.9/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/tommaso/miniconda3/envs/fractal-server-1.2.4/lib/python3.9/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/tommaso/miniconda3/envs/fractal-server-1.2.4/lib/python3.9/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/tommaso/miniconda3/envs/fractal-server-1.2.4/lib/python3.9/site-packages/starlette/middleware/cors.py", line 84, in __call__
    await self.app(scope, receive, send)
  File "/home/tommaso/miniconda3/envs/fractal-server-1.2.4/lib/python3.9/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/home/tommaso/miniconda3/envs/fractal-server-1.2.4/lib/python3.9/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/home/tommaso/miniconda3/envs/fractal-server-1.2.4/lib/python3.9/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
[...]

PR #671, opened 2023-05-08
Include extras for local-package collection (close #670)

### LINKED ISSUE #670: Task collection: package-extras not installed
I made the scMultipleX task installable by Fractal. It has an optional dependency on fractal-tasks-core. The following is in setup.cfg:
```
[options.extras_require]
fractal = 
    fractal-tasks-core >=0.9.4
```

Given that it's in a repo that's currently still private, I want to collect it via:
`fractal task collect /path/to/scmultiplex-version-details.whl --package-extras fractal`

I created the wheel with `python -m build` in the package. Then ran the command above. This successfully collects the task, but does not install the extras package.

Manually just installing the package with the extras as in `pip install ".[fractal]"` in the package folder does work though.

Do I need to do anything extra when building the wheel? Or is there a bug when using the `--package-extras fractal` option (with wheel-based collection)?
The task collection log doesn't show any warnings or errors.

### LINKED ISSUE #672: Check existence of local task package
This is currently done in common, but should be checked server-side.

Ref:
* https://github.com/fractal-analytics-platform/fractal-common/issues/25

tcompa:
* Close #670
* Close #672 (ref https://github.com/fractal-analytics-platform/fractal-common/issues/25)

To do
- [x] Fix #672
- [x] Add test(s)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,7 +62,7 @@ jobs:
         run: poetry install --with dev --no-interaction --all-extras
 
       - name: Test with pytest
-        run: poetry run coverage run -m pytest tests/test_task_api.py::test_collection_api_missing_file -s
+        run: poetry run coverage run -m pytest tests
 
       - name: Upload coverage data
         uses: actions/upload-artifact@v3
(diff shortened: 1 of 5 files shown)


tcompa:
This is currently on hold, because of #661

PR #691, opened 2023-05-12
Make SLURM-executor _completion method more robust

tcompa:


tcompa:
Current status: we are safe re: uncaught exception, but we still have a weirdly failing test.

PR #705, opened 2023-05-23
update common

ychiucco:
https://github.com/fractal-analytics-platform/fractal-common/pull/60


tcompa:
Let's bring back poetry.lock to its `main` state, and only update typing-extension - if possible.

Issue #702, opened 2023-05-19
Task-review overview


tcompa:
Me and @mfranzon tried to come up with an improved organization of tasks. Our understanding is that we are not ready to define strict access control, because we have no visibility into many-{users/teams} Fractal usage. We then propose to proceed in a way which will allow us to incrementally move towards strict access control when it will be better defined.

# TL;DR of our proposal
1. We add a `task.owner` attribute (or an attribute with a similar name), which is a standard string attribute (and not an actual DB relationship to the `user` table). This can be `None`, representing a "common" task.
2. Tasks coming from automatic collection are marked as "common", while tasks coming from the task-add endpoint are marked with some non-null `owner` value.
3. The `source` attribute of tasks coming from automatic collection is better defined, in a way which includes more metadata and then allows mutliple client-side features (e.g. https://github.com/fractal-analytics-platform/fractal/issues/345 and https://github.com/fractal-analytics-platform/fractal-web/issues/149).

# A decision to be taken

In the following, we imagine that the `task.owner` attribute is fully free, and can be specified by hand by the user adds a task - with no checks. This gives us a lot of freedom for how to evolve, but it requires the user to always specify it.

The rationale here is that users who add custom tasks are some kind of "power users", and we can ask them to accept the additional burden of always adding their nickname in a form.

If this seems too cumbersome, then there are alternative ways to proceed:
1) Automatically set it to the email address of the user (which is already known). We don't like this option, cause it exposes a private information which would be visible to all users.
2) Automatically set it to the `slurm_user` attribute of the user. We don't like this option, because it's bound to an attribute which is optional (it only becomes necessary if you actually submit a job, but a user could spend time on Fractal without even having a SLURM account.. or we could have a `local` backend and no `slurm_user` attributes).
3) We create an additional field of the `user` DB table, which is a Fractal username, and then we use it for the purpose described above (or possibly for other purposes in the future). This would be an attribute which a user has no problem with sharing (clearly different from sharing an actual email address).

# Relevant use cases

(U1) A user creates a Fractal task that is meant to be available to all users, and this task is clearly identified as "common" in Fractal (in the database and in the web client).
(U2) The same as U1 can be achieved for a Fractal-compatible Python package of tasks; in this case, Fractal takes care of: preparing a dedicated venv, installing the package, creating the correct task metadata, and adding the tasks to Fractal. Context: the Fractal admin install fractal-tasks-core or a similar task package.
[...]

PR #728, opened 2023-05-29
Add warning in export_worfklow (ref #702)

tcompa:


tcompa:
(the test was committed to main, by mistake - merging this one should fix CI in main)

PR #733, opened 2023-06-05
Allow users to edit their own tasks

### LINKED ISSUE #731: Allow users to edit their own tasks
Currently, editing tasks is restricted to superusers. I get a:

```
fractal task edit 10 --meta-file /path/to/meta_conversion.json 
ERROR:root:Server returned 403
ERROR:root:Original request: PATCH http://localhost:8011/api/v1/task/10
ERROR:root:Original payload: {"meta": {"parallelization_level": "image", "cpus_per_task": 2, "mem": 16000}}
ERROR:root:Server error message: {'detail': 'This action is restricted to superusers'}
```

That makes a lot of sense for common tasks. But user-specific tasks would be useful if they are editable by the user themselves.

ychiucco:
closes #731

### DIFF
--- a/fractal_server/app/api/v1/task.py
+++ b/fractal_server/app/api/v1/task.py
@@ -41,7 +41,6 @@ from ...db import get_db
 from ...db import get_sync_db
 from ...models import State
 from ...models import Task
-from ...security import current_active_superuser
 from ...security import current_active_user
 from ...security import User
 
@@ -362,12 +361,13 @@ async def get_task(
 async def patch_task(
     task_id: int,
     task_update: TaskUpdate,
-    user: User = Depends(current_active_superuser),
+    user: User = Depends(current_active_user),
     db: AsyncSession = Depends(get_db),
 ) -> Optional[TaskRead]:
     """
     Edit a specific task (restricted to superuser)
     """
+
     if task_update.source:
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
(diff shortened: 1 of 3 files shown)


fractal_server/app/api/v1/task.py
```diff
@@ -381,17 +381,22 @@ async def patch_task(
...
-        logger.warning(
-            f"Task owner ({db_task.owner}) differs "
-            f"from current user ({owner}). Proceed anyway."
-        )
+    if not user.is_superuser:
+        if db_task.owner is None:
+            raise HTTPException(
+                status_code=status.HTTP_403_FORBIDDEN,
+                detail=(
+                    f"Task {task_id} has no no owner: you must be a superuser"
```
tcompa, line 389: Maybe "Only a superuser can edit a task with `owner=None`"

fractal_server/app/api/v1/task.py
```diff
@@ -381,17 +381,22 @@ async def patch_task(
...
+                ),
+            )
+        else:
+            owner = user.username or user.slurm_user
+            if owner != db_task.owner:
+                raise HTTPException(
+                    status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                    detail=f"Task owner ({db_task.owner}) differs "
```
tcompa, line 397: Maybe "Current user ({owner}) cannot modify task ({task_id}) with different owner ({db_task.owner})."?

tests/test_task_api.py
```diff
@@ -502,13 +502,97 @@ async def test_post_task(client, MockCurrentUser):
         assert res.json()["owner"] == USERNAME
 
 
+async def test_patch_task_auth(
+    MockCurrentUser,
+    client,
+    task_factory,
+):
```
tcompa, line 509: Some minimal docstring would be useful

tcompa:
Looks good!

Main comment: I think I was wrong, and the correct error to raise is 403 in both cases (and not "422 or 403").

Missing:
- [ ] Update comments (both endpoint docstring and "Check match..").
- [ ] Update CHANGELOG
- [ ] A couple of comments I added to the code (typically error messages or comments)

PR #744, opened 2023-06-08
Expose `pinned_package_versions` for task collection

### LINKED ISSUE #740: Expose `pinned_package_versions` for task collection
Quoting from https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/355#issuecomment-1580543867

We explored multiple options with @mfranzon, and we don't see any which makes sense to us via conditional dependencies or something similar. We then propose that:
* fractal-tasks-core depends on a more flexible torch version (e.g. <=2.0.0)
* The sysadmin keeps installing the "correct" version (of torch, for instance) after the task collection is complete.

---

Since this is very tedious, we also propose the following workaround for doing it automatically (to be included in fractal-server - we can then open issue over there).
The `/api/v1/task/collect/pip/` currently takes this request body:
```json
{
  "package": "string",
  "package_version": "string",
  "package_extras": "string",
  "python_version": "string"
}
```
We could add an additional attribute, like `custom_package_versions`. This would be empty by default, and only at UZH we would set it to `custom_package_versions={"torch": "1.12.1"}`. The behavior of the task collection would then be:
1. Perform the whole installation in the standard way (NOTE: this must not fail!)
2. After the installation is complete, run `pip install torch==1.12.1` (where `pip` is replaced by the actual venv `pip` that is being used).

CAVEAT: this is messing with the package, and thus creating a not-so-clean log of the installation (although we would still include also the additional torch-installation logs). Such an operation is meant to be restricted to very specific cases, where there is an important dependency on hardware or system libraries - things that a regular user should not be using.

IMPORTANT NOTE 1
This workaround cannot bring us out of the versions supported by fractal-tasks-core (for instance). Say that we now require torch>=1.13.0, and then we set `custom_package_versions={"torch": "1.12.1"}`. This task-collection operation will fail, because the installation of the custom package goes conflicts with fractal-tasks-core.

IMPORTANT NOTE 2
We should never use this feature to install an additional package. For instance if the `fractal-tasks-core` does not depend on `polars`, and we specify `custom_package_versions={"polars": "1.0"}`, then task collection will fail.

MINOR NOTE:
This also fits perfectly with https://github.com/fractal-analytics-platform/fractal-server/issues/686, where we would only need to add the same `pip install` line in the script.

ychiucco:
closes #740 

Relative pr for fractal-common: https://github.com/fractal-analytics-platform/fractal-common/pull/79

### DIFF
--- a/fractal_server/tasks/collection.py
+++ b/fractal_server/tasks/collection.py
@@ -459,6 +459,9 @@ async def _pip_install(
         package_root : Path
             the location of the package manifest
     """
+
+    logger = get_logger(logger_name)
+
     pip = venv_path / "venv/bin/pip"
 
     extras = f"[{task_pkg.package_extras}]" if task_pkg.package_extras else ""
@@ -482,12 +485,16 @@ async def _pip_install(
     await execute_command(
         cwd=venv_path, command=cmd_install, logger_name=logger_name
     )
-
     if task_pkg.pinned_package_versions:
         for (
             pinned_pkg_name,
             pinned_pkg_version,
         ) in task_pkg.pinned_package_versions.items():
+
+            logger.debug(
+                "Specific version required: "
+                f"{pinned_pkg_name}=={pinned_pkg_version}"
+            )
             stdout_inspect = await execute_command(
                 cwd=venv_path,
                 command=f"{pip} show {pinned_pkg_name}",
@@ -498,7 +505,17 @@ async def _pip_install(
                 for line in stdout_inspect.split("\n")
                 if line.startswith("Version:")
             )
+            logger.debug(
+                "Preliminary check: verify that "
+                f"{pinned_pkg_version} is already installed"
+            )
             if current_version != pinned_pkg_version:
+                logger.debug(
+                    f"Currently installed version of {pinned_pkg_name}"
+                    f"({current_version}) differs from pinned version "
+                    f"({pinned_pkg_version}); "
+                    f"install version {pinned_pkg_version}."
+                )
                 await execute_command(
                     cwd=venv_path,
                     command=(
@@ -506,6 +523,11 @@ async def _pip_install(
                         f"{pinned_pkg_name}=={pinned_pkg_version}"
                     ),
                     logger_name=logger_name,
+                )
+            else:
+                logger.debug(
+                    f"Currently installed version of {pinned_pkg_name}"
+                    f"({current_version}) already matches the pinned version."
                 )
 
     # Extract package installation path from `pip show`
--- a/fractal_server/tasks/collection.py
+++ b/fractal_server/tasks/collection.py
@@ -483,6 +483,25 @@ async def _pip_install(
         cwd=venv_path, command=cmd_install, logger_name=logger_name
     )
 
+    if task_pkg.pinned_package_versions:
+        for pkg_name, pkg_version in task_pkg.pinned_package_versions.items():
+            stdout_inspect = await execute_command(
+                cwd=venv_path,
+                command=f"{pip} show {pkg_name}",
+                logger_name=logger_name,
+            )
+            current_version = next(
+                line.split()[-1]
+                for line in stdout_inspect.split("\n")
+                if line.startswith("Version:")
+            )
+            if current_version != pkg_version:
+                await execute_command(
+                    cwd=venv_path,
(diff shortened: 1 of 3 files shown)


fractal_server/tasks/collection.py
```diff
@@ -483,6 +483,25 @@ async def _pip_install(
         cwd=venv_path, command=cmd_install, logger_name=logger_name
     )
 
+    if task_pkg.pinned_package_versions:
+        for pkg_name, pkg_version in task_pkg.pinned_package_versions.items():
```
tcompa, line 487: Maybe we can use `pinned_pkg_{name,version}`, to avoid confusion with `task_pkg.package_version` (also present in this function).

tests/test_task_collection_pypi.py
```diff
@@ -158,6 +158,50 @@ async def test_pip_install(tmp_path):
...
 
 
+async def test_pip_install_pinned(tmp_path):
+    """ """
+
+    PACKAGE = "devtools"
+    VERSION = "0.8.0"
+    EXTRA = "pygments"
+    PIN_VERSION = "2.0"
+    PIN = {EXTRA: PIN_VERSION}
```
tcompa, line 168: I would add a third case
```python
PIN_INVALID = {"pydantic": "1"}
```
that should fail.

tests/test_task_collection_pypi.py
```diff
@@ -158,6 +158,50 @@ async def test_pip_install(tmp_path):
...
+
+    for pin in [None, PIN]:
```
tcompa, line 170: At first it was unclear to me why to include `None`.

Let's add a comment, like "We also include `None`, to make sure that the test with `PIN` is not passing for an arbitrary coincidence (e.g because `devtools[pygments]==0.8.0` already installed `pygments=2.0`)"

fractal_server/tasks/collection.py
```diff
@@ -482,6 +487,50 @@ async def _pip_install(
...
+                cwd=venv_path,
+                command=f"{pip} show {pinned_pkg_name}",
+                logger_name=logger_name,
+            )
+            current_version = next(
+                line.split()[-1]
+                for line in stdout_inspect.split("\n")
+                if line.startswith("Version:")
+            )
+            logger.debug(
```
tcompa, line 510: This log should go up, before the pip-show command. Otherwise it will not be visible in case pip-show fails

tests/test_task_collection_pypi.py
```diff
@@ -158,6 +159,66 @@ async def test_pip_install(tmp_path):
...
+
+    # Case 1: good pin
+    pin = {EXTRA: PIN_VERSION}
+    new_version = await _aux(venv_name="case1", pin=pin)
+    assert new_version == PIN_VERSION
+
+    # Case 2: bad pin with unexisting EXTRA version
+    pin = {EXTRA: "123456789"}
+    with pytest.raises(RuntimeError):
+        await _aux(venv_name="case2", pin=pin)
```
tcompa, line 214: Can we also verify that the error message is the expected one?

tests/test_task_collection_pypi.py
```diff
@@ -158,6 +159,66 @@ async def test_pip_install(tmp_path):
...
+
+    # Case 3: bad pin with not already installed package
+    pin = {"pydantic": "1.0.0"}
+    with pytest.raises(RuntimeError):
+        await _aux(venv_name="case3", pin=pin)
```
tcompa, line 219: Can we also verify that the error message is the expected one?

tcompa:
Thanks @ychiucco, this looks good!

A few minor details:
- Can you update the changelog?
- Can you test that `collection.log` was updated correctly? The file is in the venv path, at the same level as the `venv` folder.
- After https://github.com/fractal-analytics-platform/fractal-common/pull/79 is merged, we should make sure that the submodule here points to the fractal-common main.
- https://github.com/fractal-analytics-platform/fractal-server/pull/744/files#r1223902097
- https://github.com/fractal-analytics-platform/fractal-server/pull/744/files#r1223902908
- https://github.com/fractal-analytics-platform/fractal-server/pull/744/files#r1223902963

tcompa:
Last missing issue
- [ ] Let's try to make the test a bit faster, since it's adding almost a minute to the CI time.

PR #745, opened 2023-06-12
Delete Task endpoint

### LINKED ISSUE #738: Allow users to delete their own tasks
In #731 we implemented some preliminary version of access-control, to allow users to edit some tasks.
With the same exact logic, we can allow users to delete those same tasks:
* A superuser can delete any task
* A user can only delete a task if `task.owner` matches with (`user.username or user.slurm_user`)

What is to be added, as compared to task-editing, is that tasks could be in a relationship with one or more `WorkflowTask`s. If this is the case, then task-deletion should fail with a 422 error, and the error message should be like "Cannot remove task XY because it is currently imported in Workflows A, B and C. If you want to remove this task, then you should first remove the workflows."

ychiucco:
closes #738

### DIFF
--- a/fractal_server/app/api/v1/_aux_functions.py
+++ b/fractal_server/app/api/v1/_aux_functions.py
@@ -232,10 +232,7 @@ async def _get_task_check_owner(
         if task.owner is None:
             raise HTTPException(
                 status_code=status.HTTP_403_FORBIDDEN,
-                detail=(
-                    "Only a superuser can edit or delete a Task with "
-                    "`owner=None`."
-                ),
+                detail=("Only a superuser can get a Task with `owner=None`."),
             )
         else:
             owner = user.username or user.slurm_user
@@ -243,8 +240,8 @@ async def _get_task_check_owner(
                 raise HTTPException(
                     status_code=status.HTTP_403_FORBIDDEN,
                     detail=(
-                        f"Current user ({owner}) cannot modify task "
-                        f"({task.id}) with different owner ({task.owner})."
+                        f"Current user ({owner}) cannot get Task {task.id} "
+                        f"with different owner ({task.owner})."
                     ),
                 )
     return task
--- a/fractal_server/app/api/v1/_aux_functions.py
+++ b/fractal_server/app/api/v1/_aux_functions.py
@@ -13,8 +13,10 @@ from ...models import ApplyWorkflow
 from ...models import Dataset
 from ...models import LinkUserProject
 from ...models import Project
+from ...models import Task
 from ...models import Workflow
 from ...models import WorkflowTask
+from ...security import User
 
 
 async def _get_project_check_owner(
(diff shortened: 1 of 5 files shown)


tcompa:
Thanks @ychiucco!

One main comment about `_can_access_task`: we have a bunch of similar functions in `_aux_functions.py`. Could you move the new one there, and transform it into `_get_task_check_owner` (which should include both the ownership check and also get the task from the DB - see any of the other functions in that module). This function could raise a 403 or 404.

tcompa:
Also, can you please update the CHANGELOG?

tcompa:
> One main comment about `_can_access_task`: we have a bunch of similar functions in `_aux_functions.py`. Could you move the new one there, and transform it into `_get_task_check_owner` (which should include both the ownership check and also get the task from the DB - see any of the other functions in that module). This function could raise a 403 or 404.

And then the corresponding test would go into `test_unit_aux_functions.py`.

tcompa:
Recap:
- [ ] Move `_can_access_task` into `_aux_functions._get_task_check_owner`, and also include the handling of 404.
- [ ] Write a unit test for this new function, in `test_unit_aux_functions.py`
- [ ] Simplify the current `test_delete_task`, since it just needs to test the specific logic which is written in the endpoint function: it does not need to test the 404 and 403 cases, which are already covered in the unit test, but it does need to test the 422 case.
- [ ] Update CHANGELOG

fractal_server/app/api/v1/_aux_functions.py
```diff
@@ -210,3 +212,36 @@ async def _get_job_check_owner(
...
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Task {task_id} not found.",
+        )
+
+    if not user.is_superuser:
+        if task.owner is None:
+            raise HTTPException(
+                status_code=status.HTTP_403_FORBIDDEN,
+                detail=("Only a superuser can get a Task with `owner=None`."),
```
tcompa, line 235: I'd rather go back from "get" to "modify" here (since for the moment all users can get all tasks).

fractal_server/app/api/v1/_aux_functions.py
```diff
@@ -210,3 +212,36 @@ async def _get_job_check_owner(
...
+            )
+        else:
+            owner = user.username or user.slurm_user
+            if owner != task.owner:
+                raise HTTPException(
+                    status_code=status.HTTP_403_FORBIDDEN,
+                    detail=(
+                        f"Current user ({owner}) cannot get Task {task.id} "
```
tcompa, line 243: I'd rather go back from "get" to "modify" here (since for the moment all users can get all tasks).

tests/test_unit_aux_functions.py
```diff
@@ -281,3 +282,59 @@ async def test_get_job_check_owner(
...
+        # Test success
+        _task = await _get_task_check_owner(task_id=taskA.id, user=user, db=db)
+        assert _task.id == taskA.id
+
+        # Test fail 2: 403 FORBIDDEN
+        with pytest.raises(HTTPException) as err:
+            await _get_task_check_owner(task_id=taskB.id, user=user, db=db)
+        assert err.value.status_code == 403
+        assert err.value.detail == (
+            "Only a superuser can get a Task with `owner=None`."
```
tcompa, line 316: I'd rather go back from "get" to "modify" here (since for the moment all users can get all tasks).

tests/test_unit_aux_functions.py
```diff
@@ -281,3 +282,59 @@ async def test_get_job_check_owner(
...
+        )
+
+    async with MockCurrentUser(user_kwargs={"username": "bob"}) as user:
+        # Test fail 3: 403 FORBIDDEN
+        with pytest.raises(HTTPException) as err:
+            await _get_task_check_owner(task_id=taskA.id, user=user, db=db)
+        assert err.value.status_code == 403
+        assert err.value.detail == (
+            f"Current user ({user.username}) cannot get Task {taskA.id} "
```
tcompa, line 325: I'd rather go back from "get" to "modify" here (since for the moment all users can get all tasks).

fractal_server/app/api/v1/_aux_functions.py
```diff
@@ -210,3 +212,36 @@ async def _get_job_check_owner(
...
         )
     return dict(job=job, project=project)
+
+
+async def _get_task_check_owner(
+    *,
+    task_id: int,
+    user: User,
+    db: AsyncSession,
+) -> Task:
```
tcompa, line 222: Can we put back the original comment/docstring?
Something like
```
"""
This check constitutes a preliminary version of access control: if the current user is not a superuser and differs from the task owner (including when `owner is None`), we raise an 403 HTTP Exception.
"""
```

PR #752, opened 2023-06-15
Remove `WorkflowTask.overridden_meta`

### LINKED ISSUE #640: Import `Task.meta` into `WorkflowTask.meta`, instead of using the `WorfklowTask.overridden_meta` property
- [ ] Upon task insertion (see `Workflow.insert_task()`) into a workflow, combine the `Task.meta` (coming from the DB) with `WorkflowTaskCreate.meta` (coming from the payload, and taking priority), and store the output as `WorkflowTask.meta`.
- [ ] Replace all `WorkflowTask.overridden_meta` entries with `WorkflowTask.meta`, and remove `overridden_meta` property.

(this was part of #592, which is now closed)

ychiucco:
closes #640

### DIFF
--- a/tests/test_workflow_api.py
+++ b/tests/test_workflow_api.py
@@ -585,8 +585,7 @@ async def test_import_export_workflow(
     # original JSON file
     wf_old = WorkflowExport(**workflow_from_file).dict(exclude_none=True)
     wf_new = WorkflowExport(**workflow_exported).dict(exclude_none=True)
-    debug(wf_old)
-    debug(wf_new)
+    assert len(wf_old["task_list"]) == len(wf_new["task_list"])
     for task_old, task_new in zip(wf_old["task_list"], wf_new["task_list"]):
         assert task_old.keys() <= task_new.keys()
         for k, v in task_old.items():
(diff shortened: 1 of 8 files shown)


tests/test_workflow_api.py
```diff
@@ -581,12 +581,19 @@ async def test_import_export_workflow(
...
-    # JSON file
+    # Check that the exported workflow is an extension of the one in the
+    # original JSON file
     wf_old = WorkflowExport(**workflow_from_file).dict(exclude_none=True)
     wf_new = WorkflowExport(**workflow_exported).dict(exclude_none=True)
-
-    assert wf_old == wf_new
+    debug(wf_old)
+    debug(wf_new)
+    for task_old, task_new in zip(wf_old["task_list"], wf_new["task_list"]):
```
tcompa, line 590: This does not catch task lists of different lengths. Pleas add an assertion that their lengths are the same, before this loop.

tests/test_workflow_api.py
```diff
@@ -581,12 +581,19 @@ async def test_import_export_workflow(
...
+        assert task_old.keys() <= task_new.keys()
```
tcompa, line 590: Would the following one be equivalent?
```python
if "meta" in task_old:
    # check ...
    task_old.pop("meta")
    task_new.pop("meta")
assert task_old == task_new
```

If this also works in the same way, I find it more readable

PR #756, opened 2023-06-16
Include timestamp in workflow dir, to make it unique

### LINKED ISSUE #753: Modify `WORKFLOW_DIR` if folder already exists
If one deletes projects often (while testing) it is not hard to hit this kind error:
```
2023-06-16 08:39:47,071; ERROR; Output file /home/tommaso/Fractal/fractal-demos/examples/server/artifacts/proj_0000001_wf_0000001_job_0000001/0.metadiff.json already exists. Terminating
```

When defining `WORKFLOW_DIR`, in `submit_workflow`, we should first check if it already exists. If it does, let's slightly modify the folder name (we can simply add `_1,_2,...`, or a timestamp).

ychiucco:
closes #753

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
     * Remove `ApplyWorkflow.overwrite_input` column (\#684, \#694).
     * Make `output_dataset_id` a required argument of apply-workflow endpoint (\#681).
     * Improve error message related to out-of-space disk (\#699).
+    * Include timestamp in job working directory, to avoid name clashes (\#756).
 * Other updates to endpoints and database:
     * Add `ApplyWorkflow.end_timestamp` column (\#687, \#684).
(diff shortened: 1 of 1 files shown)


fractal_server/app/runner/__init__.py
```diff
@@ -130,25 +130,31 @@ async def submit_workflow(
...
+        )
     ).resolve()
-    if not WORKFLOW_DIR.exists():
-        original_umask = os.umask(0)
-        WORKFLOW_DIR.mkdir(parents=True, mode=0o755)
-        os.umask(original_umask)
+
+    if WORKFLOW_DIR.exists():
+        raise RuntimeError(f"Workflow dir {WORKFLOW_DIR} already exists.")
+
```
tcompa, line 144: Add comment "Create WORKFLOW_DIR with 755 permissions"

PR #757, opened 2023-06-19
Add render as batch to online method and add naming convention

### LINKED ISSUE #732: SQLite and ALTER
We have a known issue with SQLite an ALTER - see e.g. https://github.com/fractal-analytics-platform/fractal-server/issues/726.
For the moment, we kept resetting the DB table, including as part of the upcoming v1.3, but this implies that each time we hit one of those cases we have a fractal-server update that is totally disruptive (as it requires to start with a fresh db).

In principle there exist workarounds, on the SQLite side (e.g. the `render_as_batch=True`, see [link](https://stackoverflow.com/questions/30378233/sqlite-lack-of-alter-support-alembic-migration-failing-because-of-this-solutio)), but we should debug this further. We currently do have it, in the offline version of https://github.com/fractal-analytics-platform/fractal-server/blob/main/fractal_server/migrations/env.py, but it is clearly not working yet.

mfranzon:
closes #732

### DIFF
--- a/fractal_server/app/models/task.py
+++ b/fractal_server/app/models/task.py
@@ -2,7 +2,6 @@ from typing import Any
 from typing import Optional
 
 from sqlalchemy import Column
-from sqlalchemy import UniqueConstraint
 from sqlalchemy.types import JSON
 from sqlmodel import Field
 from sqlmodel import SQLModel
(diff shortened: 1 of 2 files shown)


tcompa:
Several tests were performed by @ychiucco as part of https://github.com/fractal-analytics-platform/fractal-server/pull/765.
Let's merge, and then test again as part of real-life sqlite/postgres examples.

Note: in principle we won't need to reset the schemas.

PR #775, opened 2023-06-30
Add admin Dashboard

### LINKED ISSUE #771: Experiment with `sqladmin`
EDIT: the original issue (see below) was using starlette_admin, but we later switched to https://github.com/aminalaee/sqladmin because it supports tables with multiple primary keys (like our own `LinkUserProject`).

---

Snippet that gets close to working:
```python


from starlette_admin.contrib.sqlmodel import ModelView, Admin
from fractal_server.app.models import Task, UserOAuth

# from sqlalchemy import create_engine
# engine = create_engine("sqlite:///test.db", connect_args={"check_same_thread": False})

from fractal_server.app.db import DB
from fractal_server.app.models import SQLModel

engine = DB.engine_sync()
metadata = SQLModel.metadata
metadata.create_all(engine)


# Create admin
admin = Admin(engine, title="Example: fractal-server")
admin.add_view(ModelView(Task))

app = start_application()

# Mount admin to your app
admin.mount_to(app)
```

This is fully preliminary (it's restarting the DB every time the app starts, and it also apparently does not work if we use Project instead of Task).

To be reviewed.

ychiucco:
closes #771

### DIFF
--- a/fractal_server/app/models/task.py
+++ b/fractal_server/app/models/task.py
@@ -37,7 +37,7 @@ class Task(_TaskBase, SQLModel, table=True):
     source: str = Field(unique=True)
     input_type: str
     output_type: str
-    meta: Optional[dict[str, Any]] = Field(sa_column=Column(JSON), default={})
+    metab: Optional[dict[str, Any]] = Field(sa_column=Column(JSON), default={})
     owner: Optional[str] = None
     version: Optional[str] = None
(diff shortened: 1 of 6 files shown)


tcompa:
We are currently not pursuing this path, due to:
* https://github.com/aminalaee/sqladmin/issues/539
* https://github.com/wtforms/wtforms/issues/791

PR #801, opened 2023-07-20
791 prevent having two running jobs with the same output dataset

### LINKED ISSUE #791: Prevent having two running jobs with the same output dataset
We should add one more check to the apply endpoint.

When the user submits a workflow for execution with a given output dataset, we should first check whether there exists another job with `job.status in [JobStatusType.SUBMITTED, JobStatusType.RUNNING]` and with the same `job.output_dataset`. If so, the endpoint should not create a new job and it should fail with a 422.

mfranzon:
closes #791 
~~missing test~~

### DIFF
--- a/fractal_server/app/api/v1/project.py
+++ b/fractal_server/app/api/v1/project.py
@@ -18,10 +18,10 @@ from ...db import DBSyncSession
 from ...db import get_db
 from ...db import get_sync_db
 from ...models import ApplyWorkflow
-from ...models import JobStatusType
 from ...models import ApplyWorkflowCreate
 from ...models import ApplyWorkflowRead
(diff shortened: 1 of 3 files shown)


fractal_server/app/api/v1/project.py
```diff
@@ -300,6 +300,24 @@ async def apply_workflow(
         workflow_id=workflow_id,
         **apply_workflow.dict(),
     )
+
+    stm = (
```
tcompa, line 304: I guess this check should go below the definition of `job = ApplyWorkflow(...`, right? It's true that what really matters is that the check is before the add/commit, but there is no need to create the job object if the check fails.

PR #804, opened 2023-07-24
add workflow_dump + test

### LINKED ISSUE #794: In `ApplyWorkflow`, store a copy of the `Workflow` at the time of submission
This is because the workflow may change during execution.

mfranzon:
closes #794 
remember to update `common.schemas.applyworkflow.ApplyworkflowRead`


fractal_server/app/api/v1/project.py
```diff
@@ -320,6 +320,10 @@ async def apply_workflow(
         input_dataset_id=input_dataset_id,
         output_dataset_id=output_dataset_id,
         workflow_id=workflow_id,
+        workflow_dump=dict(
+            workflow.dict(),
```
tcompa, line 324: Can we exclude `task_list` explicitly, as in `workflow.dict(exclude={"task_list"})`? It'd be to make it a bit more transparent.

fractal_server/app/api/v1/project.py
```diff
@@ -320,6 +320,10 @@ async def apply_workflow(
...
+            task_list=[wf_task.dict() for wf_task in workflow.task_list],
```
tcompa, line 325: I think this ignores `wf_task.task`, which would not be OK.

fractal_server/app/models/job.py
```diff
@@ -103,6 +108,8 @@ class ApplyWorkflow(_ApplyWorkflowBase, SQLModel, table=True):
     )
     workflow: Workflow = Relationship()
 
+    workflow_dump: dict[str, Any] = Field(sa_column=Column(JSON))
```
tcompa, line 111: Reminder: can you add the description to the docstring?

fractal_server/app/models/job.py
```diff
@@ -79,6 +81,9 @@ class ApplyWorkflow(_ApplyWorkflowBase, SQLModel, table=True):
             (Mapper attribute)
     """
 
+    class Config:
```
tcompa, line 84: I'm always confused by this. Is it because of `Column(JSON)`?

tcompa:
Reminders:
* Update CHANGELOG
* Fix https://github.com/fractal-analytics-platform/fractal-common/issues/84 and pull the update common in this branch

fractal_server/app/models/job.py
mfranzon: exactly, not a standard type for sqlalchemy https://docs.sqlalchemy.org/en/20/core/type_basics.html#generic-camelcase-types

fractal_server/app/api/v1/project.py
mfranzon: true, I'll add in the next commit

fractal_server/app/models/job.py
tcompa: Thanks!

fractal_server/app/api/v1/project.py
mfranzon: ok

fractal_server/app/models/job.py
mfranzon: sure

fractal_server/app/api/v1/project.py
```diff
@@ -320,6 +320,13 @@ async def apply_workflow(
         input_dataset_id=input_dataset_id,
         output_dataset_id=output_dataset_id,
         workflow_id=workflow_id,
+        workflow_dump=dict(
+            workflow.dict(exclude={"task_list"}),
+            task_list=[
+                dict(wf_task.task.dict(), task=wf_task.dict())
```
tcompa, line 326: Same as above: what about excluding `task`? Maybe it's a bit pedantic, but it's to make it clear that it won't be duplicated.

mfranzon: :+1:

PR #814, opened 2023-08-01
add docs_info and docs_link

### LINKED ISSUE #549: Add `Task.docs_info` and `Task.docs_link`
This would be an optional field, that would mostly be present for tasks collected from fractal-tasks-core package.
The MIP task, for instance, would have this value: "https://fractal-analytics-platform.github.io/fractal-tasks-core/api_files/fractal_tasks_core.cellpose_segmentation.html#fractal_tasks_core.cellpose_segmentation.cellpose_segmentation".

The nice consequence is that when browsing tasks in the web client we would be only one click away from the docs.

An obvious requirement is that docstrings of the tasks are detailed and precise.

mfranzon:
closes #549


tcompa:
Thanks @mfranzon!

Quick question (just as a clarification): is it expected that the `Optional[HttpUrl]` `docs_link` field in the model is transformed in a nullable `AutoString()` column in the migration file? I guess it's all fine, but I simply wanted to double check.

Other than that, this is ready to merge.

PR #816, opened 2023-08-22
Fix pyyaml and docker-compose dependencies

### LINKED ISSUE #813: Unpin pyyaml version
In https://github.com/fractal-analytics-platform/fractal-server/pull/812 we are pinning pyyaml to 5.3.1 (downgrading from 5.4.1), to avoid problems due to the recent cython=3 release:

* https://github.com/yaml/pyyaml/issues/601
(notice the endless list of issues referencing to this one)
* https://github.com/yaml/pyyaml/issues/724
* https://github.com/yaml/pyyaml/issues/736
* https://github.com/yaml/pyyaml/issues/728

Luckily we only use this dependency in the dev/docs groups, so that it doesn't really matter for our package. Still let's unpin it (in the future) as soon as a solution is found.

For the record:
* https://github.com/fractal-analytics-platform/fractal-server/security/dependabot/14

tcompa:
Close #813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Documentation:
     * Update documentation configuration and GitHub actions (\#811).
 * Tests:
     * Move `tests/test_common.py` into `fractal-common` repository (\#808).
+    * Switch to `docker compose` v2 and unpin `pyyaml` version (\#816).
 
 # 1.3.4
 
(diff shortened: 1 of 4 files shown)


tcompa:
CI failure with pytest-docker v2 -> to check.

Meanwhile, this PR would also have the nice effect of removing quite a few (dev) dependencies, which never hurts ;)

tcompa:
The main issue was related to dash/underscore in the container name, see https://docs.docker.com/compose/migrate/#service-container-names.

Issue #854, opened 2023-09-19
Handle non-Fractal-compatible tasks more robustly


tcompa:
I am playing with a dummy task, with `command="echo"`, which clearly does not write any metadiff output.
This results into
```python traceback
INFO:     127.0.0.1:41412 - "POST /api/v1/project/1/workflow/1/apply/?input_dataset_id=1&output_dataset_id=2 HTTP/1.1" 202 Accepted
2023-09-19 15:30:57,406 - WF1_job1 - INFO - Start execution of workflow "zxc"; more logs at /tmp/artifacts/proj_0000001_wf_0000001_job_0000001_20230919_133057/workflow.log
2023-09-19 15:30:57,413 - WF1_job1 - INFO - Workflow "zxc" failed (unkwnon error).
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/tommaso/Fractal/fractal-web/lib/fractal-server/venv/lib/python3.10/site-packages/fractal_server/app/runner/__init__.py", line 228, in submit_workflow
    output_dataset_meta = await process_workflow(
  File "/home/tommaso/Fractal/fractal-web/lib/fractal-server/venv/lib/python3.10/site-packages/fractal_server/app/runner/_local/__init__.py", line 163, in process_workflow
    output_dataset_metadata = await async_wrap(_process_workflow)(
  File "/home/tommaso/Fractal/fractal-web/lib/fractal-server/venv/lib/python3.10/site-packages/fractal_server/app/runner/common.py", line 244, in async_wrapper
    return await loop.run_in_executor(executor, pfunc)
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/tommaso/Fractal/fractal-web/lib/fractal-server/venv/lib/python3.10/site-packages/fractal_server/app/runner/_local/__init__.py", line 56, in _process_workflow
    output_task_pars = execute_tasks(
  File "/home/tommaso/Fractal/fractal-web/lib/fractal-server/venv/lib/python3.10/site-packages/fractal_server/app/runner/_common.py", line 605, in execute_tasks
    current_task_pars = this_wftask_future.result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 458, in result
    return self.__get_result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
[...]

Issue #649, opened 2023-04-27
Review db connections in API and background tasks


tcompa:
Ref https://github.com/fractal-analytics-platform/fractal-server/issues/647

1. Why is the apply-workflow endpoint using both sync/async session? Can we simplify it? Do we need an additional close after refresh? If we keep two sessions, both must be closed (and the sync one is currently not closed!).
2. In submit_workflow (which takes place in background), the `db_sync.commit` statement before the long workflow execution is not fully closing the session. This is visible via the presence of a non-trivial identity map. Adding an explicit `db_sync.close` seems to work, and then the session is re-opened upon `merge`. **Shall we move towards context-managers? Probably that's the best way to go.**

Point 1 is a bug. Point 2 is a bit unclear, but for sure it requires some update.

tcompa:
For the record: the following test passes, which (unexpectedly) shows that the `new_db_sync.commit()` is not enough  to fully clear a DB connection.

```python
async def test_sync_db_close(db_sync, db):
    """
    This test is used as an example to show that the commit() method is not
    enough to fully close a db_sync connection.
    """

    from fractal_server.app.models.task import Task
    from fractal_server.app.db import DB

    db.add(
        Task(
            name="mytask",
            input_type="image",
            output_type="zarr",
            command="cmd",
            source="/source",
        )
    )
    await db.commit()
    await db.close()

    new_db_sync = next(DB.get_sync_db())
    task = new_db_sync.get(Task, 1)
    debug(task)
    task.name = "new_name"
    new_db_sync.merge(task)
    new_db_sync.commit()
    debug(new_db_sync.identity_map)
    assert new_db_sync.identity_map
    new_db_sync.close()
    debug(new_db_sync.identity_map)
    assert not new_db_sync.identity_map
```

tcompa:
Also, for the record:
> Context manager use is optional; otherwise, the returned [Session](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session) object may be closed explicitly via the [Session.close()](https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.Session.close) method. Using a try:/finally: block is optional, however will ensure that the close takes place even if there are database errors:
> (https://docs.sqlalchemy.org/en/14/orm/session_api.html#sqlalchemy.orm.sessionmaker)

tcompa:
Ref (related to "ghost" connections that seems to always be present):
[...]

Issue #647, opened 2023-04-24
Database locked after unknown issue


jluethi:
The database on the UZH deployment using server 1.2.2 became locked today. After Adrian ran into an error when providing a wrong json file during adding a task to a workflow (likely an error in the meta json file), the database became locked and any user using fractal client would see this error:
```
(fractal-client-1.2.0) atschan@pelkmanslab-slurm-master-001:~/fractal-demos/examples/09_add_custom_cellpose$ fractal project list
Traceback (most recent call last):
  File "/data/homes/atschan/.conda/envs/fractal-client-1.2.0/lib/python3.9/site-packages/httpcore/backends/asyncio.py", line 34, in read
    return await self._stream.receive(max_bytes=max_bytes)
  File "/data/homes/atschan/.conda/envs/fractal-client-1.2.0/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 1265, in receive
    await self._protocol.read_event.wait()
  File "/data/homes/atschan/.conda/envs/fractal-client-1.2.0/lib/python3.9/asyncio/locks.py", line 226, in wait
    await fut
asyncio.exceptions.CancelledError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/data/homes/atschan/.conda/envs/fractal-client-1.2.0/lib/python3.9/site-packages/httpcore/_exceptions.py", line 10, in map_exceptions
    yield
  File "/data/homes/atschan/.conda/envs/fractal-client-1.2.0/lib/python3.9/site-packages/httpcore/backends/asyncio.py", line 36, in read
    return b""
  File "/data/homes/atschan/.conda/envs/fractal-client-1.2.0/lib/python3.9/site-packages/anyio/_core/_tasks.py", line 118, in __exit__
    raise TimeoutError
TimeoutError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/data/homes/atschan/.conda/envs/fractal-client-1.2.0/lib/python3.9/site-packages/httpx/_transports/default.py", line 60, in map_httpcore_exceptions
    yield
  File "/data/homes/atschan/.conda/envs/fractal-client-1.2.0/lib/python3.9/site-packages/httpx/_transports/default.py", line 353, in handle_async_request
    resp = await self._pool.handle_async_request(req)
[...]

PR #860, opened 2023-09-21
Fix sqlite "database is locked" error

zonia3000:
## Checklist before merging

This solves the `sqlite3.OperationalError: database is locked` that appears sometimes when using sqlite.

Reference: https://www.sqlite.org/wal.html

> WAL provides more concurrency as readers do not block writers and a writer does not block readers. Reading and writing can proceed concurrently.


tcompa:
Thanks Sonia! We'll review this asap.

Possibly related to multiple past issues:
* https://github.com/fractal-analytics-platform/fractal-server/issues/647
* https://github.com/fractal-analytics-platform/fractal-server/issues/854
* https://github.com/fractal-analytics-platform/fractal-server/issues/649

tcompa:
For the record, the CI also passes locally - where I use sqlite (while the GitHub action uses postgres).

mfranzon:
I think we can merge this PR. 

Locally I have no problem, CI works fine with sqlite. 

Looking at the Disadvantages section in the [docs](https://sqlite.org/wal.html), seems that we are safe. The only thing I want to mention i point 2 
> All processes using a database must be on the same host computer; WAL does not work over a network filesystem.

Which can be a potential problem if we want to use sqlite on a network (fractal-containers) which could be easily solved via docker volumes. BUT at the moment we are happy with postgres so, the sqlite usage is strictly related to local testing.

mfranzon (approved):

tcompa:
Thanks Marco for the review. Maybe what we can add (after #891 is merged) is a test of `set_db` which always runs with SQLite (even when the postgres extra is installed), so that we would be also testing these new lines.

tcompa:
> Which can be a potential problem if we want to use sqlite on a network (fractal-containers) which could be easily solved via docker volumes. BUT at the moment we are happy with postgres

Agreed, not an issue.

> the sqlite usage is strictly related to local testing.

Local testing or local demo-ing. But we always print a very clear warning saying the SQLite is not fully supported.

tcompa:
> Thanks Marco for the review. Maybe what we can add (after #891 is merged) is a test of `set_db` which always runs with SQLite (even when the postgres extra is installed), so that we would be also testing these new lines.

#891 is on hold, therefore this test should be included within the current setup.

tcompa:
> therefore this test should be included within the current setup.

Implemented with #915. If the CI passes, this is ready to merge

tcompa:
Before #915:
> 57.14% of new lines are covered.

After #915:
> 100% of new lines are covered.

Thanks @zonia3000 and @ychiucco! I'm now merging this one.

PR #867, opened 2023-09-27
Add check in PATCH Workflow

### LINKED ISSUE #858: PATCH workflow lets user have two workflows with the same name
POST endpoint has logic:
```python
    await _check_workflow_exists(name=workflow.name, project_id=project_id, db=db)
```
to check that there is no existing workflow for the same project and with the same name.

PATCH endpoint does not use this function, and then lets a user break the constraint mentioned above.

(note that also workflow-import POST endpoint has this check in-place)

---

If this is a strict constraint, then we should also add it to the PATCH endpoint.

ychiucco:
closes #858 

## Checklist before merging


tcompa:
Looks good, thanks @ychiucco!

PR #880, opened 2023-10-02
Dependencies injection

### LINKED ISSUE #856: Review use of dependency-injection module
The following script
```python
from syringe import Inject
from devtools import debug

debug(Inject)

def foo():
    return 52

x = Inject(foo)
debug(x)
debug(Inject._dependencies)
```
outputs
```
syringe.py:61 _Inject.__init__
    'INIT' (str) len=4
x.py:4 <module>
    Inject: <syringe._Inject object at 0x7fd841228b50> (_Inject)
syringe.py:79 _Inject.__call__
    'CALL' (str) len=4
    _callable: <function foo at 0x7fd83f9a5a20> (function)
x.py:11 <module>
    x: 52 (int)
x.py:12 <module>
    Inject._dependencies: {} (dict) len=0
```

It looks to me and @ychiucco that this means that we are not using `Inject` at all!
Let's make sure that this is the case, and evaluate whether it's worth keeping this module (and what was its motivation in the first place).

ychiucco:
closes #856 

## Checklist before merging


tcompa:
Closing in favor of #891

Issue #620, opened 2023-04-14
Refactor logging to integrate uvicorn/gunicorn, fastapi, and fractal-server


tcompa:
Ref https://github.com/fractal-analytics-platform/fractal-server/pull/619#issuecomment-1508122395

tcompa:
Ref:
* https://docs.python.org/3/howto/logging-cookbook.html#deploying-web-applications-using-gunicorn-and-uwsgi
* https://docs.python.org/3/howto/logging-cookbook.html#running-a-logging-socket-listener-in-production

PR #891, opened 2023-10-04
Review monkeypatching of `settings`

ychiucco:
## Checklist before merging


tcompa:
CI is passing locally, but taking very long:
```
Results (1028.62s):
       300 passed
         2 xfailed
        30 warnings
```

fractal_server/config.py
```diff
@@ -419,4 +419,7 @@ def check(self):
 
 
 def get_settings(settings=Settings()) -> Settings:
+    logging.warn("Fractal Settings:")
```
tcompa, line 422: This function is deprecated - see the warning message appearing in the CI.

fractal_server/config.py
```diff
@@ -419,4 +419,7 @@ def check(self):
...
+    for k, v in settings.dict().items():
+        logging.warn(f"{k}: {v}")
```
tcompa, line 424: Maybe we should replace `v` with `"***"` when `k` includes strings like `"PASSWORD"` or `"SECRET"`?

This is because, until we tackle #620, it is not 100% clear where logs are ending up - since it depends on which startup command is used.

fractal_server/tasks/collection.py
```diff
@@ -206,15 +205,17 @@ def create_package_dir_pip(
...
+
     user = FRACTAL_PUBLIC_TASK_SUBDIR
     if task_pkg.package_version is None:
         raise ValueError(
             f"Cannot create venv folder for package `{task_pkg.package}` "
             "with `version=None`."
         )
     package_dir = f"{task_pkg.package}{task_pkg.package_version}"
-    venv_path = settings.FRACTAL_TASKS_DIR / user / package_dir  # type: ignore
+    venv_path = Path(settings.FRACTAL_TASKS_DIR) / user / package_dir
```
tcompa, line 218: Why is this `Path` needed?

tests/fixtures_server.py
```diff
@@ -123,51 +123,69 @@ def get_patched_settings(temp_path: Path):
     return settings
 
 
+modules_to_patch = [
```
tcompa, line 126: This list is a nice improvement on the hard-coding of all module names.
Two comments:
1. Please add an explanatory comment, or turn it into a function with an informative docstring. In the current form, it's hard to guess what's happening.
2. The explanation should also say why we exclude modules that contain "migrations".
3. Is there a reason for not checking that "get_settings" is in the file? It would cost a bit more (you would need to read the text file) but it would also reduce the number of patched functions. And it would maybe also make the goal of this block more transparent. Let's maybe first check which fraction of modules do have a match.

tests/fixtures_server.py
```diff
@@ -123,51 +123,69 @@ def get_patched_settings(temp_path: Path):
...
+    )
+    if "migrations" not in module
+]
+
+# SETTINGS fixtures
+
+
 @pytest.fixture(scope="session", autouse=True)
-def override_settings(tmp777_session_path):
+async def set_default_test_settings(tmp777_session_path):
```
tcompa, line 138: Please add an informative docstring here.

tests/fixtures_server.py
```diff
@@ -81,7 +81,7 @@ def check_python_has_venv(python_path: str, temp_path: Path):
         )
 
 
-def get_patched_settings(temp_path: Path):
+def get_default_test_settings(temp_path: Path):
```
tcompa, line 84: Missing return type, and maybe also a (brief) docstring.

tests/fixtures_server.py
tcompa, line 126: Maybe we can capitalize it as in `MODULES_TO_PATCH`? This is sometimes our (informal) way of denoting variables which are harcoded or computed in the main block of a module (i.e. outside functions/classes).

tests/fixtures_server.py
```diff
@@ -123,51 +123,69 @@ def get_patched_settings(temp_path: Path):
...
+    tmp_path = tmp777_session_path("server_folder")
+    patched_settings = get_default_test_settings(tmp_path)
 
-    Inject.override(get_settings, _get_settings)
-    try:
-        yield settings
-    finally:
-        Inject.pop(get_settings)
+    for k, v in request.param.items():
+        setattr(patched_settings, k, v)
```
tcompa, line 160: `pydantic` does not automatically perform checks upon `setattr` (unless we change its the model's Config block).
Maybe it's worth double-checking the model validity after this loop of updates, with an explicit call of the model init (or any other pydantic method for validation).

tests/fixtures_server.py
```diff
@@ -123,51 +123,69 @@ def get_patched_settings(temp_path: Path):
...
+                raising=False,
+            )
+        yield
 
-    settings = get_patched_settings(tmp_path)
 
-    def _get_settings():
-        return settings
+@pytest.fixture
+async def override_settings_startup(tmp777_session_path, monkeypatch, request):
```
tcompa, line 152: A docstring here would be useful.
Let's make sure that the next PR on this issue is easier to write ;)

tcompa, line 152: Is there a reason why this is async?

tests/test_fixtures_settings_factory.py
```diff
@@ -1,27 +0,0 @@
-"""
```
tcompa, line 1: Should we re-add one or two of these tests - as a kind of documentation/example of the fixtures?

tests/test_task_collection_api.py
```diff
@@ -82,7 +80,10 @@ async def test_collection(
         * if called twice, the same tasks are returned without installing
     """
 
-    override_settings_factory(FRACTAL_TASKS_DIR=(tmp_path / "test_collection"))
+    override_settings_runtime(
+        FRACTAL_TASKS_DIR=str(tmp_path / "test_collection")
```
tcompa, line 84: Is this a string or a Path?

tests/test_fixtures_settings_factory.py
tcompa: EDIT: I just found them.

tests/fixtures_server.py
```diff
@@ -123,51 +123,69 @@ def get_patched_settings(temp_path: Path):
...
+    if not request.__dict__.get("param"):
```
tcompa, line 153: Can we avoid relying on the private attribute `__dict__`?

tests/fixtures_server.py
tcompa: > Is there a reason for not checking that "get_settings" is in the file? It would cost a bit more (you would need to read the text file) but it would also reduce the number of patched functions. And it would maybe also make the goal of this block more transparent. Let's maybe first check which fraction of modules do have a match.

From Yuri: yes, it's much faster to try the patch and handle the error. And this is also done automatically by pytest, via `raising=False` (see docs here: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.MonkeyPatch.setattr).

tests/test_unit_fixtures_settings.py
```diff
@@ -0,0 +1,79 @@
...
+
+    debug(DEFAULT_SETTINGS)
+
+    assert DEFAULT_SETTINGS.DB_ENGINE == "sqlite"
+    assert not DEFAULT_SETTINGS.SQLITE_PATH
+
+    settings = get_settings()
+    debug(settings)
+
+    if settings.DB_ENGINE == "sqlite":
```
tcompa, line 21: I like this test.
Can we make it more general? That is, it should always have an assertion, not only when using sqlite.

tests/test_unit_fixtures_settings.py
```diff
@@ -0,0 +1,79 @@
...
+        assert settings.SQLITE_PATH
+
+
+@pytest.mark.parametrize(
+    "override_settings_startup",
+    [{"JWT_SECRET_KEY": "FooBar"}],
+    indirect=True,
+)
+def test_override_settings_startup_sigle(override_settings_startup):
```
tcompa, line 30: Typo in function name

tests/test_unit_fixtures_settings.py
ychiucco: The idea is: if it's not `sqlite`, than it must be `postgres`. So we are implicitly asserting that `DEFAULT_SETTINGS.DB_ENGINE == "postgres"`.
Making it explicit requires zero effort though.

tcompa: > The idea is: if it's not sqlite, than it must be postgres. So we are implicitly asserting that DEFAULT_SETTINGS.DB_ENGINE == "postgres".
> Making it explicit requires zero effort though.

OK, it makes sense. Having it explicit wouldn't hurt though, simply to avoid having this conversation again in a month.

tcompa:
After multiple reviews/discussions with @ychiucco and @mfranzon, this is currently on hold.
The current state is not yet working, as displayed in `test_override_settings_runtime`:
```python
def test_override_settings_runtime(override_settings_runtime):

    from fractal_server.config import get_settings

    startup_settings = get_settings()
    startup_key = startup_settings.JWT_SECRET_KEY

    new_secret_key = startup_key + "-FANCY-TAIL"
    override_settings_runtime(JWT_SECRET_KEY=new_secret_key)

    new_settings = get_settings()

    assert startup_key == "secret_key"  # from `get_default_test_settings`
    assert new_settings.JWT_SECRET_KEY == "secret_key-FANCY-TAIL"

    # NOTE: pytest first executes `override_settings_runtime`.
    # In our case, this means running the necessary code in order to build
    # `new_secret_key`.
    # After that, the test is run again from the beginning with overriden
    # settings, and that's why `startup_settings.JWT_SECRET_KEY` is
    # "secret_key-FANCY-TAIL" and not  "secret_key".
    assert startup_settings.JWT_SECRET_KEY == "secret_key-FANCY-TAIL"
```

Note: during work on this issue, it is always useful to run the CI locally with sqlite, rather than postgres. In at least one case, this showed different errors which are not reproduced with postgres.

PR #896, opened 2023-10-06
Log request body in endpoints

### LINKED ISSUE #861: All/many endpoints should log their request body (at the DEBUG level)
In this way, by simply switching to `FRACTAL_LOGGING_LEVEL=0` we immediately get a very detailed view of what is happening.

ychiucco:
closes #861 

## Checklist before merging

### DIFF
--- a/fractal_server/app/api/v1/dataset.py
+++ b/fractal_server/app/api/v1/dataset.py
@@ -1,4 +1,5 @@
 import json
+import logging
 from pathlib import Path
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


fractal_server/app/api/v1/dataset.py
```diff
@@ -50,6 +51,9 @@ async def create_dataset(
     """
     Add new dataset to current project
     """
+    logging.debug(f"POST /project/{project_id}/dataset")
```
tcompa, line 54: This is not needed, since we already have logs about the URL of the endpoint. What we need to add is information about the request body.

tcompa:
Closing as stale.

I'm not convinced that the added complexity of a middleware is worth the current goal (having a verbose debug mode, which prints all request bodies). Let's proceed with a naive implementation (that is, a `logger.debug` statement in each endpoint).

PR #897, opened 2023-10-09
Remove default Dataset from Project creation

### LINKED ISSUE #674: Remove automatic creation of default dataset when creating a project
The project-creation endpoint also takes care of creating a first dataset. I think this feature gives us little benefit, since the dataset is anyway empty and it will have to be handled again later (e.g. to change its name/type or add resources). Also: this could be the input dataset for a workflow, but the output dataset still needs to be created.

In the same spirit of https://github.com/fractal-analytics-platform/fractal-server/issues/566, I propose to remove this feature. That is, a new project would be created without any dataset. The goal is to remove not-so-crucial branches, in order to:
* Simplify maintenance and documentation
* Make the addition of new features more transparent, instead of relying on some "hidden" feature.

Higher-level side note: there were good reasons for adding this kind of features in the first place, but IMO this is not the case any more after a few months of experience.

Any feedback? @jluethi @gusqgm

ychiucco:
closes #674 

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 # Unreleased
 
 * Do not automatically create a dataset upon project creation (\#897).
+* Remove `ProjectCreate.default_dataset_name` attribute (\#897).
 
 # 1.3.11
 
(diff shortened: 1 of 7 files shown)


tcompa:
I also cleaned up the situation of the project/job/dataset API tests a bit, since most of them were in `test_project_api.py`, and in some cases some features were tested multiple times.

tcompa:
Ready to merge, thanks @ychiucco!

PR #898, opened 2023-10-09
838 move datasetmetahistory into datasethistory

### LINKED ISSUE #826: Fully deprecate `HISTORY_LEGACY`
As part of #803, the `Dataset.meta["history"]` property corresponds to the new version, while the legacy version is still stored in `Dataset.meta["HISTORY_LEGACY"]`. We should later fully deprecate the `HISTORY_LEGACY` one.

Ref:
* https://github.com/fractal-analytics-platform/fractal/issues/543

### LINKED ISSUE #827: Review/deprecate custom history handling in `/project/{project_id}/job/{job_id}` endpoint
The current code is an oversimplified version of the new endpoints introduced via #803, which doesn't fit well with the run-a-workflow-through-different-jobs use case (because it would only return the history of the current job).

We should either review this feature, or (once we verify that the new endpoints work as expect) fully deprecate it.

### LINKED ISSUE #838: Move `Dataset.meta["history"]` into `Dataset.history`
This is to be re-assessed after a bit more real-life testing and improvement of the new history (ref #803).

Reasons for moving to a separate DB column:

1. We could better define its structure, on the line of
```python
# Option 0
DatasetHistoryItemType = dict[Literal["worfklowtask", "status", "parallelization"]]
history: list[DatasetHistoryItemType]

# Option 1
class DatasetHistoryItem(Basemodel):
    worfklowtask: WorkflowTaskRead   # TBD
    status: WorkflowTaskStatusType
    parallelization: dict  # TBD
```

2. Better scope separation between `Dataset.meta` and `Dataset.history`:
    * We should avoid "leaking" the full history of a dataset to the task (which we are currently doing)
    * We could differentiate the current priority rules which are in-place for `meta` in some endpoints but in principle only apply to `history`

mfranzon:
## Checklist before merging

close #827 
close #838
close #826


fractal_server/app/models/dataset.py
```diff
@@ -58,6 +58,7 @@ class Dataset(_DatasetBase, SQLModel, table=True):
         }
     )
     meta: dict[str, Any] = Field(sa_column=Column(JSON), default={})
+    history: dict[str, Any] = Field(sa_column=Column(JSON), default={})
```
tcompa, line 61: This should be a list

fractal_server/app/api/v1/dataset.py
```diff
@@ -432,7 +432,7 @@ async def get_workflowtask_status(
 
     # Lowest priority: read status from DB, which corresponds to jobs that are
     # not running
-    history = dataset.meta.get("history", [])
+    history = dataset.history or []  # meta.get("history", [])
```
tcompa, line 435: Can history be `None`? If not, this check is not necessary.

fractal_server/app/runner/__init__.py
```diff
@@ -249,15 +249,18 @@ async def submit_workflow(
...
-        new_meta = {}
-        for key, value in output_dataset_meta.items():
-            if key != "history":
-                # For non-history keys, replace with new value
-                new_meta[key] = value
-            else:
-                # For history key, append to existing entry
-                new_meta[key] = output_dataset.meta.get(key, []) + value
-        output_dataset.meta = new_meta
+        output_dataset.history = output_dataset_meta.pop("history")
```
tcompa, line 252: This doesn't sound fully correct, at least in terms of variable name.s

What is the output of `process_workflow`?
1. If it's a instance of `meta`, then we can keep calling it `output_dataset_meta`. But then it should not include a `history` key.
2. If it's a collection of both the new meta and history, then let's make it something like `output_whatever=dict(meta=..., history=...).
3. Another option is that `process_workflow` returns a tuple, with meta and history.

I'd be in favor of option 2, but all of them are OK.

tests/test_history.py
```diff
@@ -75,9 +75,9 @@ async def test_get_workflowtask_status(
             RESULTS[status].add(ID)
 
         # Create output_dataset and job
-        meta = dict(history=history)
+        history = history
```
tcompa, line 78: `history = history`? :)

tcompa:
Something more: can we now fully remove `history` from the `ApplyWorkflowRead` model?

Higher-level: I am not a fan of adding custom attributes to the `Read` models that do not correspond to a db-table column. If special information are needed, we can always use class properties or methods, or we can create an additional Pydantic model that inherits e.g. from `ApplyWorkflowRead` and which is used in a specific endpoint. The `Read` models should be used in all the CRUD endpoints, and they should be as close as possible the the db contents.

tests/test_history.py
mfranzon: Just to be sure that history==history XD

tcompa:
With [4bee986](https://github.com/fractal-analytics-platform/fractal-server/pull/898/commits/4bee9860f797742e851be2b951622d9782b4ed49), I'm adding input validation to dummy tasks that we use for tests.

This is to be closer to the way fractal-tasks-core work, but also to catch errors like the one fixed with [4c9c911](https://github.com/fractal-analytics-platform/fractal-server/pull/898/commits/4c9c911755f00bcc1f040dd5b6f4ff31a9bafee8), which currently went undetected and only appeared as part of a fractal-client CI (https://github.com/fractal-analytics-platform/fractal-client/actions/runs/6496628972/job/17643989673), in this form:
```
        'log': (
            'TASK ERROR:Task id: 1 (dummy), e.workflow_task_order=0\n'
            'TRACEBACK:\n'
            'Traceback (most recent call last):\n'
            '  File "/tmp/pytest-of-runner/pytest-0/test_workflow_apply0/FRACTAL_TASKS_DIR/.fractal/fractal_tasks_core'
            '_alpha0.0.1a0/venv/lib/python3.11/site-packages/fractal_tasks_core_alpha/tasks/dummy.py", line 149, in <m'
            'odule>\n'
            '    metadata_update = dummy(**pars)\n'
            '                      ^^^^^^^^^^^^^\n'
            "TypeError: dummy() got an unexpected keyword argument 'history'\n"
        ),
```

tcompa:
With [ee88dba](https://github.com/fractal-analytics-platform/fractal-server/pull/898/commits/ee88dba893bc4eef9a56bd30ca4bd8287f25f6de), I tried to use `_DatasetHistoryItem` in the runner, but I found some issues because I would need actual `WorkflowTaskRead` objects in there (with all their primary/foreign keys), and in the tests we use Mock objects. For the moment I reverted the commit.

tcompa:
> With [ee88dba](https://github.com/fractal-analytics-platform/fractal-server/pull/898/commits/ee88dba893bc4eef9a56bd30ca4bd8287f25f6de), I tried to use `_DatasetHistoryItem` in the runner, but I found some issues because I would need actual `WorkflowTaskRead` objects in there (with all their primary/foreign keys), and in the tests we use Mock objects. For the moment I reverted the commit.

I'm still trying to evaluate whether we are using `_DatasetHistoryItem` correctly.
1. We have it in the init file of the local/slurm runners, but it's only a type hint.
2. We have it in the dataset schemas, but it's not very useful for `DatasetCreate` and `DatasetUpdate`.
3. It is indeed relevant for the `DatasetRead` model, but this is a very small scope.

I think we should use it a bit more, and especially when constructing the history.

tcompa:
> I'm still trying to evaluate whether we are using `_DatasetHistoryItem` correctly.

After discussing with @mfranzon, we opted for removing this class from the runner, where it was currently only used as a slightly wrong type hint. See https://github.com/fractal-analytics-platform/fractal-server/pull/898/commits/3135fec78acd256b9381b4efa3a16f209563d661

tcompa:
> @mfranzon should we really include `DatasetUpdate.history`? I cannot see an obvious use case.
> 
> We can also decide to include it, to have a more consistent API, and then add a check somewhere (either in a model validator or within the PATCH endpoint) so that it is not actually allowed.

We agreed that consistency is important, and we'll keep that attribute. We'll raise a 422 from the endpoint if a user tries to modify it.

tcompa:
The last thing I want to check is the behavior for when we start from a dataset with a non-empty history (e.g. because we are continuing a workflow execution). Maybe this is already part of a test, or I will add one.

Other than that, I think this is ready!

PR #917, opened 2023-10-18
Make `test_commands` stateless with postgres

### LINKED ISSUE #913: `test_commands` is not stateless when using postgres
The `db` fixture makes tests stateless, w.r.t. dropping all tables.
In the `test_commands.py` module we call fractalctl commands via subprocess, and we do not revert this operation at the end.
This is OK for sqlite, since the db files are function-scoped, but not for postgres tests.

ychiucco:
closes #913 

## Checklist before merging

### DIFF
--- a/tests/fixtures_server.py
+++ b/tests/fixtures_server.py
@@ -11,8 +11,11 @@ Institute for Biomedical Research and Pelkmans Lab from the University of
 Zurich.
 """
 import logging
+import os
 import random
+import shlex
 import shutil
+import subprocess
 from dataclasses import dataclass
 from dataclasses import field
 from pathlib import Path
@@ -30,6 +33,7 @@ from httpx import AsyncClient
 from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 
+import fractal_server
 from fractal_server.app.db import get_db
 from fractal_server.config import get_settings
 from fractal_server.config import Settings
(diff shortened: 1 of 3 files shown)


tests/fixtures_server.py
```diff
@@ -166,6 +170,80 @@ def _get_settings():
...
+    )
+    debug(res.stdout)
+    debug(res.stderr)
+    debug(res.returncode)
+    assert res.returncode == 0
+
+    yield
+
+    if DB_ENGINE == "postgres":
+        cmd = "poetry run alembic downgrade base"
```
tcompa, line 232: Please add comment pointing to https://alembic.sqlalchemy.org/en/latest/tutorial.html#downgrading

tcompa: (latest -> 14)

tests/fixtures_server.py
```diff
@@ -166,6 +170,80 @@ def _get_settings():
             Inject.override(get_settings, get_settings_orig[0])
 
 
+@pytest.fixture(scope="function")
+async def prepare_config_and_db(tmp_path):
```
tcompa, line 174: Move back to original module

tests/fixtures_server.py
```diff
@@ -166,6 +170,80 @@ def _get_settings():
...
+            encoding="utf-8",
+            capture_output=True,
+            cwd=FRACTAL_SERVER_DIR,
+        )
+        debug(res.stdout)
+        debug(res.stderr)
+        debug(res.returncode)
+        assert res.returncode == 0
+
+    Path.unlink(FRACTAL_SERVER_DIR / ".fractal_server.env")
```
tcompa, line 244: Comment like "Remove config file, to make test stateless")

tests/fixtures_server.py
tcompa: Also say that we only do it for postgres because when using sqlite it may be useful to quickly look at the file.

tcompa:
Thanks @ychiucco!

PR #927, opened 2023-10-25
Review `ApplyWorkflow` table definition

### LINKED ISSUE #667: Review job table to allow deleting datasets or workflows
A job is in relationship with
* a project
* a workflow
* one or two datasets

Expected behaviors:
1. Removing a project removes all associated jobs (ref #633, in progress with #666)
2. Removing a workflow removes all associated jobs - to be implemented.
3. Removing a dataset removes all associated jobs - to be implemented.

Note that 2 and 3 can be counterintuitive, let's make sure we are OK with them. If not, we can find a workaround.

Current situation is not acceptable, because removing a dataset would lead to a broken state in which the job table has rows with foreign keys pointing to non-existing datasets.

### LINKED ISSUE #928: Deferred db objects in `submit_workflow`

### LINKED ISSUE #929: Review project deletion
Once we manage to implement https://github.com/fractal-analytics-platform/fractal-server/issues/667#issuecomment-1776830647, let's do the same for `ApplyWorkflow.project_id`. That is, deleting a project should set the `ApplyWorkflow.project_id` FK to `null` for all jobs of that project.

A couple more additions would be needed:
1. We need to keep track of the user who called the apply endpoint. This should be easily achieved by adding a `ApplyWorkflow.user_dump: str` field, and the tiny bit of logic that sets it to `user.email` in the apply endpoint.
2. `ApplyWorkflow` access-control in endpoints should also check whether `project_id` is `null`. If so, only a superuser should have access.

Ref
* #667 
* #928

ychiucco:
closes #667
closes #928 
closes #929 

## Checklist before merging


fractal_server/app/models/job.py
```diff
@@ -87,35 +86,56 @@ class Config:
...
+            back_populates="list_jobs_output",
             lazy="selectin",
             primaryjoin="ApplyWorkflow.output_dataset_id==Dataset.id",
         )
     )
-    workflow: Workflow = Relationship()
 
-    workflow_dump: Optional[dict[str, Any]] = Field(sa_column=Column(JSON))
+    user_dump: dict[str, Any] = Field(
+        sa_column=Column(JSON, nullable=False, server_default="{}")
```
tcompa, line 119: I think we don't want a server default here

tcompa: (this also applies to the other "dump" attributes)

fractal_server/app/models/job.py
```diff
@@ -87,35 +86,56 @@ class Config:
...
+    working_dir: Optional[str]
+    working_dir_user: Optional[str]
+    first_task_index: int
+    last_task_index: int
 
     start_timestamp: datetime = Field(
         default_factory=get_timestamp,
-        nullable=False,
-        sa_column=Column(DateTime(timezone=True)),
+        sa_column=Column(DateTime(timezone=True), nullable=False),
```
tcompa, line 138: Just for the record: was this needed for some specific reason?

fractal_server/app/runner/__init__.py
```diff
@@ -187,6 +187,15 @@ async def submit_workflow(
         job.status = JobStatusType.RUNNING
         db_sync.merge(job)
         db_sync.commit()
+        # After Session.commit() is called, either explicitly or when using a
```
tcompa, line 190: Let's add a link to our issue #928

fractal_server/migrations/versions/eebd9f68ad58_add_dumps_to_applyworkflow.py
```diff
@@ -0,0 +1,98 @@
...
+            sa.Column(
+                "output_dataset_dump",
+                sa.JSON(),
+                server_default="{}",
+                nullable=False,
+            )
+        )
+        batch_op.alter_column(
+            "workflow_dump",
+            existing_type=postgresql.JSON(astext_type=sa.Text()),
```
tcompa, line 45: I'm confused by this kind of `alter_column` arguments. Are they autogenerated by alembic? Why do they differ from what is above? And most importantly: why is there a reference to postgresql at all?

fractal_server/app/models/job.py
tcompa: > Can we split the migration in two migrations (or two blocks of the same migration.. it's the same)? The first adds the column with a server default, so that it can correctly upgrade an existing DB. The second one removes the default.

I now see that this is done, in the migration, but in a way which I do not fully understand (see comment below, related to the `alter_table` parameters).

tcompa:
Let's also fix #929.
Missing:
> ApplyWorkflow access-control in endpoints should also check whether project_id is null. If so, only a superuser should have access.

fractal_server/app/api/v1/project.py
```diff
@@ -302,6 +302,9 @@ async def apply_workflow(
         input_dataset_id=input_dataset_id,
         output_dataset_id=output_dataset_id,
         workflow_id=workflow_id,
+        user_dump=user.dict(),
```
tcompa, line 305: Let's only store the user email, as per #929

fractal_server/app/models/workflow.py
```diff
@@ -114,6 +115,7 @@ class Workflow(_WorkflowBase, SQLModel, table=True):
             cascade="all, delete-orphan",
         ),
     )
+    job_list: list["ApplyWorkflow"] = Relationship(back_populates="workflow")
```
tcompa, line 118: No need for double quotes here, right?

fractal_server/app/models/job.py
```diff
@@ -87,35 +86,56 @@ class Config:
...
+    project: Optional["Project"] = Relationship(  # noqa: F821
+        back_populates="job_list"
+    )
+
+    workflow_id: Optional[int] = Field(foreign_key="workflow.id")
+    workflow: Optional["Workflow"] = Relationship(  # noqa: F821
+        back_populates="job_list"
+    )
+
+    input_dataset_id: Optional[int] = Field(foreign_key="dataset.id")
```
tcompa, line 100: I think `input_dataset` and `output_dataset` should also be optional, like `workflow`.

fractal_server/app/models/job.py
```diff
@@ -87,35 +86,56 @@ class Config:
...
-    first_task_index: int
-    last_task_index: int
 
+    project_id: Optional[int] = Field(foreign_key="project.id")
+    project: Optional["Project"] = Relationship(  # noqa: F821
+        back_populates="job_list"
+    )
+
+    workflow_id: Optional[int] = Field(foreign_key="workflow.id")
+    workflow: Optional["Workflow"] = Relationship(  # noqa: F821
```
tcompa, line 96: Why can we import Dataset but not Workflow? Unless there are special reasons, let's keep it homogeneous.

tests/test_workflow_api.py
```diff
@@ -969,7 +969,7 @@ async def test_delete_workflow_failure(
     """
     GIVEN a Workflow in a relationship with a Job
     WHEN we try to DELETE that Workflow
-    THEN we fail with a 422
+    THEN we succeed
```
tcompa, line 972: This test is not working correctly: it tests the deletion of a workflow which is not related to any job)

tcompa: Also, to be added: after deletion, we should assert that job has `workflow_id` equal to `None`. And we should also assert something on the `job.workflow` attribute.

tests/test_unit_db_models.py
```diff
@@ -328,3 +330,62 @@ async def test_insert_task_with_meta_none(
...
+        debug(job)
+
+        new_dataset = await db.get(Dataset, dataset.id)
+        assert len(new_dataset.list_jobs_input) == 1
+        assert len(new_dataset.list_jobs_output) == 1
+
+        new_job_1 = await db.get(ApplyWorkflow, job.id)
+        debug(new_job_1)
+        assert new_job_1.input_dataset_id == 42
+        assert new_job_1.output_dataset_id == 42
```
tcompa, line 382: What should we assert about `new_job_1.input_dataset` at this point?

tests/test_unit_db_models.py
```diff
@@ -328,3 +330,62 @@ async def test_insert_task_with_meta_none(
...
+
+        await db.delete(new_dataset)
+        await db.commit()
+        await db.close()
+
+        new_job_2 = await db.get(ApplyWorkflow, job.id)
+        debug(new_job_2)
+        assert new_job_2.input_dataset_id is None
```
tcompa, line 390: What should we assert about `new_job_1.input_dataset` at this point?

tests/test_project_api.py
```diff
@@ -205,10 +205,13 @@ async def test_delete_project(
         )
 
         # Check that a project-related job exists
-        stm = select(ApplyWorkflow).join(Project).where(Project.id == p["id"])
-        res = await db.execute(stm)
-        jobs = list(res)
-        assert len(jobs) == 1
+        stm = select(ApplyWorkflow).where(ApplyWorkflow.project_id == p["id"])
```
tcompa, line 208: We should also test `project.job_list` here, and verify that it points to the same (number of) jobs.

tests/test_project_api.py
```diff
@@ -226,9 +229,9 @@ async def test_delete_project(
...
-        jobs = await db.execute(stm)
-        jobs = list(jobs)
-        debug(jobs)
-        assert len(jobs) == 0
+        # Assert that total number of jobs is still 1, but without project_id
+        await db.refresh(job)
+        assert not job.project_id
+        assert not job.input_dataset_id
+        assert not job.output_dataset_id
+        assert not job.workflow_id
```
tcompa, line 237: I would generally use `something is None` rather than `not something`. It's better defined (what if `workflow_id=0`?) and it's more readable.

fractal_server/migrations/versions/eebd9f68ad58_add_dumps_to_applyworkflow.py
```diff
@@ -0,0 +1,98 @@
+"""Add dumps to ApplyWorkflow
+
+Revision ID: eebd9f68ad58
+Revises: 99ea79d9e5d2
+Create Date: 2023-10-25 12:46:29.503628
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
```
tcompa, line 10: I think this import really should not be here, unless I'm missing something.

tcompa:
I pushed 1698825702a371ee795d417d67b7143c56931fba, since I think that a job should not include the whole project as an attribute.

Based on the same logic, I would then explore whether the same is possible for `applyworkflow.workflow`, `applyworkflow.input_dataset` and `applyworkflow.output_dataset`.

fractal_server/app/models/job.py
tcompa: Actually we may even remove this relationship, if this does not create other problems.

fractal_server/migrations/versions/84bf0fffde30_add_dumps_to_applyworkflow.py
```diff
@@ -0,0 +1,88 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("applyworkflow", schema=None) as batch_op:
+        batch_op.add_column(
+            sa.Column(
+                "user_dump", sa.String(), server_default="", nullable=False
```
tcompa, line 23: Replace `""` with `"__UNDEFINED__"`

fractal_server/app/schemas/project.py
```diff
@@ -47,6 +48,7 @@ class ProjectRead(_ProjectBase):
 
     id: int
     dataset_list: list[DatasetRead] = []
+    workflow_list: list[WorkflowRead] = []
```
tcompa, line 51: Add `job_list`

tests/test_commands.py
```diff
@@ -229,3 +229,24 @@ def test_migrations_on_old_data_sqlite(tmp_path: Path, testdata_path: Path):
...
+    values = cur.execute("SELECT * FROM applyworkflow").fetchone()
+    columns = [desc[0] for desc in cur.description]
+    data_dict = {
+        k: v
+        for k, v in zip(columns, values)
+        if (v is not None) and (k not in ["id", "user_dump"])
+    }
+    with pytest.raises(sqlite3.IntegrityError):
+        cur.execute(
+            f"INSERT INTO applyworkflow ({str(list(data_dict.keys()))[1:-1]}) "
```
tcompa, line 245: Make more readable please

fractal_server/app/schemas/project.py
```diff
@@ -49,6 +50,7 @@ class ProjectRead(_ProjectBase):
     id: int
     dataset_list: list[DatasetRead] = []
     workflow_list: list[WorkflowRead] = []
+    job_list: list[ApplyWorkflowRead] = []
```
tcompa, line 53: Is this tested anywhere?

fractal_server/migrations/versions/84bf0fffde30_add_dumps_to_applyworkflow.py
```diff
@@ -0,0 +1,91 @@
...
+        batch_op.add_column(
+            sa.Column(
+                "output_dataset_dump",
+                sa.JSON(),
+                server_default="{}",
+                nullable=False,
+            )
+        )
+        batch_op.alter_column(
+            "workflow_dump", existing_type=sa.JSON(), nullable=False
```
tcompa, line 46: Missing test:
* Start (before this migration) with a row with `ApplyWorkflow.workflow_dump=None`
* Apply the migration

tcompa: If this fails, let's add a `server_default` also here (if it exists in `alter_column`)

fractal_server/migrations/versions/84bf0fffde30_add_dumps_to_applyworkflow.py
```diff
@@ -0,0 +1,91 @@
...
+            "input_dataset_id", existing_type=sa.INTEGER(), nullable=True
+        )
+        batch_op.alter_column(
+            "output_dataset_id", existing_type=sa.INTEGER(), nullable=True
+        )
+
+    with op.batch_alter_table("applyworkflow", schema=None) as batch_op:
+        batch_op.alter_column("user_dump", server_default=None)
+        batch_op.alter_column("input_dataset_dump", server_default=None)
+        batch_op.alter_column("output_dataset_dump", server_default=None)
```
tcompa, line 64: If needed, let's replicate this line for `workflow_dump`

tests/data/test-1.3.11.db
ychiucco, line 1:
```sql
UPDATE applyworkflow SET workflow_dump = null;
```

fractal_server/app/models/job.py
```diff
@@ -87,34 +85,33 @@ class Config:
...
+    )
+    output_dataset_dump: dict[str, Any] = Field(
+        sa_column=Column(JSON, nullable=False)
     )
-    output_dataset: Dataset = Relationship(
-        sa_relationship_kwargs=dict(
-            lazy="selectin",
-            primaryjoin="ApplyWorkflow.output_dataset_id==Dataset.id",
-        )
+    workflow_dump: dict[str, Any] = Field(
```
tcompa, line 103: Bring this back to Optional, and include nullable=True within Column

fractal_server/migrations/versions/84bf0fffde30_add_dumps_to_applyworkflow.py
```diff
@@ -0,0 +1,112 @@
...
+        batch_op.add_column(
+            sa.Column(
+                "output_dataset_dump",
+                sa.JSON(),
+                server_default="{}",
+                nullable=False,
+            )
+        )
+
+        applyworkflow = table("applyworkflow", column("workflow_dump"))
```
tcompa, line 48: Remove whole workflow_dump block

fractal_server/migrations/versions/84bf0fffde30_add_dumps_to_applyworkflow.py
```diff
@@ -0,0 +1,112 @@
...
+            "input_dataset_id", existing_type=sa.INTEGER(), nullable=False
+        )
+        batch_op.alter_column(
+            "workflow_id", existing_type=sa.INTEGER(), nullable=False
+        )
+        batch_op.alter_column(
+            "project_id", existing_type=sa.INTEGER(), nullable=False
+        )
+
+        batch_op.alter_column(
```
tcompa, line 98: Remove whole workflow_dump block

fractal_server/app/api/v1/project.py
```diff
@@ -302,6 +302,9 @@ async def apply_workflow(
         input_dataset_id=input_dataset_id,
         output_dataset_id=output_dataset_id,
         workflow_id=workflow_id,
+        user_dump=user.email,
```
tcompa, line 305: Rename with `user_email`

fractal_server/app/models/job.py
```diff
@@ -87,34 +85,33 @@ class Config:
...
-            primaryjoin="ApplyWorkflow.input_dataset_id==Dataset.id",
-        )
+    project_id: Optional[int] = Field(foreign_key="project.id")
+
+    workflow_id: Optional[int] = Field(foreign_key="workflow.id")
+
+    input_dataset_id: Optional[int] = Field(foreign_key="dataset.id")
+    output_dataset_id: Optional[int] = Field(foreign_key="dataset.id")
+
+    user_dump: str = Field(sa_column=Column(String, nullable=False))
```
tcompa, line 96: Do not use `sa_column`. That is, `Field(nullable=False)`

fractal_server/app/models/dataset.py
```diff
@@ -37,6 +37,19 @@ class Dataset(_DatasetBase, SQLModel, table=True):
 
     id: Optional[int] = Field(default=None, primary_key=True)
     project_id: int = Field(foreign_key="project.id")
+
+    list_jobs_input: list["ApplyWorkflow"] = Relationship(  # noqa: F821
```
tcompa, line 41: Can we import `ApplyWorkflow` here?

fractal_server/app/models/dataset.py
```diff
@@ -37,6 +37,19 @@ class Dataset(_DatasetBase, SQLModel, table=True):
...
+        sa_relationship_kwargs=dict(
+            lazy="selectin",
+            primaryjoin="ApplyWorkflow.input_dataset_id==Dataset.id",
+        )
+    )
+    list_jobs_output: list["ApplyWorkflow"] = Relationship(  # noqa: F821
```
tcompa, line 47: Can we import `ApplyWorkflow` here?

tcompa:
Update `ApplyWorkflowRead` (missing attributes)

tests/test_dataset_api.py
```diff
@@ -167,21 +166,35 @@ async def test_delete_dataset_failure(
...
 
-        # Test successful dataset deletion
         res = await client.delete(
-            f"api/v1/project/{project.id}/dataset/{dummy_ds.id}"
+            f"api/v1/project/{project.id}/dataset/{input_ds.id}"
         )
         assert res.status_code == 204
 
+        await db.refresh(job)
+        assert not job.input_dataset_id
```
tcompa, line 186: Replace with `is None`

tests/test_dataset_api.py
```diff
@@ -167,21 +166,35 @@ async def test_delete_dataset_failure(
...
+        assert job.output_dataset_id
```
tcompa, line 187: Replace with `is not None`

tests/test_dataset_api.py
```diff
@@ -167,21 +166,35 @@ async def test_delete_dataset_failure(
...
 
-        # Check that you cannot delete datasets in relationship with a job
-        for ds_id in (input_ds.id, output_ds.id):
-            res = await client.delete(
-                f"api/v1/project/{project.id}/dataset/{ds_id}"
-            )
-            assert res.status_code == 422
-            assert f"still linked to job {job.id}" in res.json()["detail"]
+        # Assert that `Dataset.list_jobs_*` are correctly populated
```
tcompa, line 172: Test that link is correct, in terms of IDs.

tests/test_unit_db_models.py
```diff
@@ -328,3 +330,65 @@ async def test_insert_task_with_meta_none(
         wf = Workflow(name="my wfl", project_id=project.id)
         args = dict(arg="test arg")
         await wf.insert_task(t0.id, db=db, args=args)
+
+
+async def test_delete_dataset_cascade(db, MockCurrentUser, project_factory):
```
tcompa, line 335: Remove this test, after moving whatever is needed to one of the `test_dataset_api.py` tests

tests/test_workflow_api.py
```diff
@@ -968,38 +968,38 @@ async def test_delete_workflow_failure(
...
+        assert job.workflow_id == workflow.id
+
         res = await client.delete(
-            f"api/v1/project/{project.id}/workflow/{workflow_2.id}"
+            f"api/v1/project/{project.id}/workflow/{workflow.id}"
         )
         assert res.status_code == 204
 
+        await db.refresh(job)
+        assert not job.workflow_id
```
tcompa, line 1001: Use `is None`

tcompa (approved):

PR #953, opened 2023-11-10
Expose `redirect_url` for OAuth clients

tcompa:
Close #952

## Checklist before merging
- [x] I added an appropriate entry to `CHANGELOG.md`

PR #974, opened 2023-11-16
Review use of `Enum`s

tcompa:
Close #953 

## Checklist before merging

### DIFF
--- a/fractal_server/app/models/job.py
+++ b/fractal_server/app/models/job.py
@@ -90,5 +90,5 @@ class ApplyWorkflow(_ApplyWorkflowBase, SQLModel, table=True):
     end_timestamp: Optional[datetime] = Field(
         default=None, sa_column=Column(DateTime(timezone=True))
     )
-    status: JobStatusType = JobStatusType.SUBMITTED
+    status: str = JobStatusType.SUBMITTED
     log: Optional[str] = None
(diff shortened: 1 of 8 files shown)


fractal_server/app/models/job.py
```diff
@@ -115,5 +92,10 @@ class Config:
     end_timestamp: Optional[datetime] = Field(
         default=None, sa_column=Column(DateTime(timezone=True))
     )
-    status: JobStatusType = JobStatusType.SUBMITTED
+    status: str = JobStatusType.SUBMITTED
     log: Optional[str] = None
+
+    @validator("status")
```
tcompa, line 98: @mfranzon let's decide whether this validator is actually valuable

tcompa:
I'm pretty convinced that the bulk of the PR is fine, just removing all `Enum`s from `app/models`, and only keeping them within `app/schemas`. I'm left with a doubt about whether to include a validator in `ApplyWorkflow` - cc @mfranzon

fractal_server/app/models/job.py
mfranzon: We can relax the validator as long as there is no user interaction (PATCH) with the job_status

mfranzon (approved):

fractal_server/app/models/job.py
tcompa: Agreed. It doesn't add so much, apart from complexity. Now removed with https://github.com/fractal-analytics-platform/fractal-server/pull/974/commits/83160692fee6a872a411e3c25f0772409ce6ffe8.

PR #1013, opened 2023-11-21
Patch current user

### LINKED ISSUE #964: Expose endpoint for user to modify (some of) their own attributes
The goal is that a user is always authorized to make a call to `PATCH ../me`, which is currently not possible.

Note that there must still be strict constraints on the payload, which can only include:
* `cache_dir`
* `password`
* `slurm_accounts`

and no other attributes.

### LINKED ISSUE #1014: Review current-user endpoints
(related to #1013)

* Remove both GET and PATCH for `/auth/users/me/` (provided by fastapi-users).
* Rename `GET /auth/whoami/` into `GET /auth/current-user/`
* Rename the upcoming unrestricted patch-me endpoint into `PATCH /auth/current-user/`. Note that this one will use a strict model for what can be updated - ref #964

Note:
* Every endpoint under `/auth/users/` remains superuser-restricted
* Every endpoint under `/auth/current-user/` is accessible to any user, since it only provides access to the current user

### LINKED ISSUE #1015: Introduce first version of `validate_password`
The logic can be extremely simple (e.g. minimum length of 4 characters and maximum length of 80), for the moment.
The goal is to introduce this overloaded function, and test it in all relevant places:
1. user registration
2. user update via `/auth/users/{id}`
3. user update via new patch endpoint (ref #964)

ychiucco:
closes #964 
closes #1014 
closes #1015

## Checklist before merging


fractal_server/app/routes/auth.py
```diff
@@ -53,10 +56,25 @@
...
-async def whoami(
-    user: User = Depends(current_active_user),
+@router_auth.patch("/me", response_model=UserRead)
+async def patch_current_user(
+    user_update: UserUpdate,
+    current_user: User = Depends(current_active_user),
+    user_manager=Depends(get_user_manager),
 ):
+    updates = {
+        k: getattr(user_update, k)
```
tcompa, line 66: # Comment 1

As of recent PRs, I suggest we try to _not_ be smart, and proceed with the simplest possible code.
Here we are using 5 lines, a for loop, a `getattr` function, a second (identical) `getattr` function, the construction of a list object, and a conditional.
An alternative option would be
```python
updates = {}
if user_update.cache_dir is not None:
    updates["cache_dir"] =  user_update.cache_dir
if user_update.password is not None:
    updates["password"] =  user_update.password
```
or am I missing something?

# Comment 2
The current solution does not raise any error if the client tries to modify other attributes. I propose that we define a more explicit Pydantic schema (e.g. `UserUpdateStrict`) which only includes the attributes that can be modified.
I propose that we are even more strict

tcompa:
Let's also review the code of `update_me` in https://github.com/fastapi-users/fastapi-users/blob/master/fastapi_users/router/users.py#L89.
I think we should provide an endpoint which is consistent with the others (e.g. the register-user one). That is, I'd rather include the same exact password checks.

tcompa:
Addendum: we don't currently overload the `validate_password` method of the user manager, but we may do it in the future (actually it would be good to get started, by at least requiring a minimum password length). That's why I'd rather have a consistent handling of passwords across different endpoints.

fractal_server/app/routes/auth.py
```diff
@@ -53,10 +56,25 @@
...
+        for k in ["cache_dir", "password"]  # "slurm_accouts" to be added
+        if getattr(user_update, k) is not None
+    }
+
+    return await user_manager.update(
+        UserUpdate(**updates), current_user, safe=True
```
tcompa, line 72: Let's keep safe=True here, I agree.

tests/test_security.py
```diff
@@ -114,11 +114,21 @@ async def test_edit_user(registered_client, registered_superuser_client):
...
     assert res.status_code == 403
 
-    # PATCH/me with non-superuser user
-    res = await registered_client.patch(
+    # PATCH /users/me do not exists
+    res = await registered_superuser_client.patch(
         f"{PREFIX}/users/me/", json={"slurm_user": "asd"}
     )
-    assert res.status_code == 403
+    assert res.status_code == 404
```
tcompa, line 121: Great, let's do it also for the GET

fractal_server/app/routes/auth.py
```diff
@@ -56,24 +62,18 @@
 )
 
 
-@router_auth.patch("/me", response_model=UserRead)
+@router_auth.patch("/current-user", response_model=UserRead)
```
tcompa, line 65: Let's start adding trailing slash to our custom endpoints, even if they would be added automatically at the end of this file. This will make things more readable, as one would immediately see the correct path.

fractal_server/app/routes/auth.py
```diff
@@ -56,24 +62,18 @@
...
-    }
-
     return await user_manager.update(
-        UserUpdate(**updates), current_user, safe=True
+        UserUpdate(**user_update.dict()), current_user, safe=True
     )
 
 
-@router_auth.get("/whoami", response_model=UserRead)
+@router_auth.get("/current-user", response_model=UserRead)
```
tcompa, line 76: Same as above, regarding trailing slash.

fractal_server/app/schemas/user.py
```diff
@@ -53,6 +54,19 @@ class UserUpdate(schemas.BaseUserUpdate):
...
 
+class UserUpdateStrict(BaseModel):
+    """
+    Attributes that every user can self-edit
+    """
+
+    cache_dir: Optional[str]
+    password: Optional[str]
+
+    _cache_dir = validator("cache_dir", allow_reuse=True)(
```
tcompa, line 65: Question: what happens for `UserUpdateStrict(password="1234")`? is this validator only acting when `cache_dir` is set?

fractal_server/app/routes/auth.py
```diff
@@ -43,8 +44,16 @@
 # Include users routes, after removing DELETE endpoint (ref
 # https://github.com/fastapi-users/fastapi-users/discussions/606)
 users_router = fastapi_users.get_users_router(UserRead, UserUpdate)
```
tcompa, line 46: Let's clarify the comment, to mention that we are removing some `/me` endpoints to replace them with our own endpoints below.

fractal_server/app/routes/auth.py
```diff
@@ -53,17 +70,45 @@
...
+        )
+    except exceptions.InvalidPasswordException as e:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail={
+                "code": ErrorCode.UPDATE_USER_INVALID_PASSWORD,
+                "reason": e.reason,
+            },
+        )
+    except exceptions.UserAlreadyExists:
```
tcompa, line 94: I don't think we can ever enter this branch, and in that case I would not include it.

That's only my first guess, but it's easy to check: if we can have test coverage for these lines, then let's keep them.

fractal_server/app/schemas/user.py
```diff
@@ -52,6 +55,34 @@ class UserUpdate(schemas.BaseUserUpdate):
         val_absolute_path("cache_dir")
     )
 
+    @root_validator(pre=True)
+    def cant_set_none(cls, values):
```
tcompa, line 59: Use `validator` instead of `root_validator`, on the line of:
```python
    @validator("is_superuser", "is_superuser_2", always=True)
    def cant_set_none(cls, v):
        if v is None:
            raise ValueError("Cannot set xxx=None")
        return v
```

tcompa:
To do: check all use cases as in https://github.com/fractal-analytics-platform/fractal-server/issues/1014#issuecomment-1822555227

tests/test_security.py
```diff
@@ -99,43 +100,193 @@ async def test_show_user(registered_client, registered_superuser_client):
...
+        f"{PREFIX}/current-user/", json={"cache_dir": "not_abs"}
+    )
+    assert res.status_code == 422
+    # - String attribute '{attribute}' cannot be None
+    res = await registered_client.patch(
+        f"{PREFIX}/current-user/", json={"cache_dir": None}
+    )
+    assert res.status_code == 422
+
+    # PASSWORD
```
tcompa, line 138: split cache_dir and password tests, in view of upcoming PR about password update

tests/test_security.py
```diff
@@ -99,43 +100,193 @@ async def test_show_user(registered_client, registered_superuser_client):
...
+    res = await registered_client.patch(
+        f"{PREFIX}/current-user/", json={"password": NEW_PASSWORD}
+    )
+    assert res.status_code == 200
 
-    # PATCH/{user_id} with superuser
+    from httpx import AsyncClient
+    from asgi_lifespan import LifespanManager
+
+    async with AsyncClient(
```
tcompa, line 170: Can we use the `client` fixture directly?

fractal_server/app/routes/auth.py
```diff
@@ -53,17 +70,51 @@
...
+        user = await user_manager.update(update, current_user, safe=True)
+    except exceptions.InvalidPasswordException as e:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail={
+                "code": ErrorCode.UPDATE_USER_INVALID_PASSWORD,
+                "reason": e.reason,
+            },
+        )
+    except ValueError as e:
```
tcompa, line 100: Let's remove this except

fractal_server/app/schemas/user.py
```diff
@@ -49,7 +51,34 @@ class UserUpdate(schemas.BaseUserUpdate):
...
     _cache_dir = validator("cache_dir", allow_reuse=True)(
-        val_absolute_path("cache_dir")
+        val_absolute_path("cache_dir", accept_none=True)
+    )
+
+    @validator(
+        "is_active",
+        "is_verified",
+        "is_superuser",
+        "email",
```
tcompa, line 61: Add always=False

tcompa:
Let's make thing simple, and enforce the following constraint on both `PATCH /auth/users/{user_id}/` and `PATCH /auth/current-user/`:
```
payload: {"cache_dir": null} --> 422 (independently on the computational backend)
```

fractal_server/app/schemas/_validators.py
```diff
@@ -44,14 +44,19 @@ def val(integer: int):
     return val
 
 
-def val_absolute_path(attribute: str):
+def val_absolute_path(attribute: str, accept_none: bool = False):
```
tcompa, line 47: We are adding the `accept_none`  here in `val_absolute_path`, but when we use `val_absolute_path` we never set `accept_none=True`.

Maybe this is a leftover from previous versions in this PR?

fractal_server/app/schemas/_validators.py
```diff
@@ -10,7 +10,7 @@ def valstr(attribute: str, accept_none: bool = False):
     """
 
     def val(string: str):
-        if string is None:
+        if (string is None) or (string == ""):
```
tcompa, line 13: This cannot be like this, since `accept_none` has the side effect of accepting an empty string.

fractal_server/app/schemas/_validators.py
tcompa: I removed this argument.

fractal_server/app/schemas/_validators.py
tcompa: I reverted this update, and brought back `test_task_update` to its `main` version. There is not reason why this PR would need to modify the expected behavior of `TaskUpdate`.

PR #1021, opened 2023-11-24
Review MockCurrentUser

### LINKED ISSUE #1016: Review use of `MockCurrentUser` fixture
We often use a `with MockCurrentUser(..)` block, and then call an anonymous `client` from within the block.

Question 1: how does the anonymous client get authorized? Likely this is due to re-using the cached session token.

Question 2: is the cached session token removed in the `__atexit__` method? If not, we should fix this.

ychiucco:
closes #1016 

## Checklist before merging


tcompa:
Closing for now, cause we got blocked and the PR went stale. It would make more sense to start afresh in the future, after reviewing the problem.

Issue #1066, opened 2023-12-04
Update to Pydantic v2


tcompa:
Requires:
* https://github.com/fractal-analytics-platform/fractal-server/issues/1065

Note that there is no urgency for the pydantic v2 update, and it will also come with some subtle choices (e.g. possibly about optional attributes of request-body schemas). However I expect it will be much easier to make this v2 update here in fractal-server, and learn valuable lessons for when we'll have to do it in fractal-tasks-core (ref https://github.com/fractal-analytics-platform/fractal-tasks-core/issues/592).

PR #1067, opened 2023-12-04
Add `slurm_accounts` attribute to Users

### LINKED ISSUE #956: Expose SLURM accounts via API
# Current status

Currently, there exists several different ways of choosing a specific SLURM account for job execution. From the more general to the more specific one:

1. The admin may set a default account in the SLURM/Fractal configuration file, which applies by default to all jobs (unless it is overridden). This looks like
```json
{
    "default_slurm_config": {
        "partition": "main",
        "cpus_per_task": 1,
        "account": "somecommonaccount"
    }
}
```
2. The user may add this information to the `WORKER_INIT` variable, as in
```python
WORKER_INIT = "#SBATCH --account=myaccount"
```
3. For a given task, a user can add a new `meta` property, with the key-value pair `account->something`.

Each one of these three options is then transformed in the same kind of line within the sbatch submission script.

# Future changes

There are multiple possible extensions of this mechanism:

4. We may add a column to the user DB table, that lists all their available slurm accounts (similar to the `slurm_user` attribute).
    * Why this is useful: this information can be used to offer a pre-filled drop-down menu in fractal-web, at the job-run stage
    * Why this is not useful: At a difference from `slurm_user`, a Fractal user should be able to freely add/remove/modify their own accounts (since this does not require any sudo privilege). But modifying a user is currently restricted to superusers, and then this wouldn't immediately work. We can still make it work, but we need to rework the permissions of the user CRUD endpoints.
5. We may include a string query parameter for the apply endpoint, named `slurm_account`, and offer the corresponding input form in fractal-web. Inference of possible values would not be available from fractal-server, but we could in principle build it within fractal-web (based on the previous values that were used by a given user).
6. Independently on how the apply-workflow endpoint receives this information (be it one of a list of user's accounts or a free-field endpoint query parameter), further changes are needed:
    * The DB model for jobs must be modified, as it should also include a `slurm_account` optional column.
    * The attribute `job.slurm_account` should then be used in the backend runner component (or neglected, for the local runner), and it should be passed as an additional argument to the `_slurm_submit_setup` function, which (if not `None`) will take priority over any existing value of `account` (e.g. the ones included through methods 1 and 3 above).
    * There will always be a chance of conflicts, if the user sets an account both via approach 2 above and any other method (1, 3, or any other future method). Note that this is already an issue in the current status. We could add a check that makes the job execution fail in the presence of multiple values for `account` (or we may verify what is the SLURM behavior in this case, as I guess that this will already be an error).

A couple of general notes:
* We will (likely) add some backend-specific information to DB model and API endpoints. This is not the first time, and it's acceptable because SLURM is our main target, and because these attributes are clearly optional. Let's keep it in mind that this approach won't scale forever, especially if we introduce a new runner.
* Within SLURM, we may be moving towards _four_ possible sources of the same information (actually five: task.meta, workflowtask.meta, job.worker_init, SLURM configuration file, and potentially one more). I'd be in favor of pruning some branch, since this situation could lead to unclear situations. Some options:
    * We start filtering some strings from worker-init (e.g. `# SBATCH --account`), and returning a 422 error from the workflow-apply endpoint in case they are present (together with a suggestion of where this should go).

Feedback is welcome.

ychiucco:
closes #956 

## Checklist before merging


fractal_server/app/schemas/user.py
```diff
@@ -93,6 +96,7 @@ class UserCreate(schemas.BaseUserCreate):
     slurm_user: Optional[str]
     cache_dir: Optional[str]
     username: Optional[str]
+    slurm_accounts: Optional[list[str]] = []
```
tcompa, line 99: I don't think this is correct.
1. `Optional` means that the type can be `None`, but this is not a valid use case.
2. I think using mutable objects as default is not recommended for class attributes either (apart from being plain wrong for function arguments). Isn't there some better way, possibly based on `Field`?

fractal_server/main.py
```diff
@@ -138,6 +139,8 @@ async def _create_first_user(
                     )
                     if slurm_user:
                         kwargs["slurm_user"] = slurm_user
+                    if slurm_accounts:
```
tcompa, line 142: Let's not use this pattern (`if something` instead of `if something is not None`).

tests/test_unit_schemas.py
```diff
@@ -341,13 +341,27 @@ def test_user_create():
...
     with pytest.raises(ValidationError):
-        u = UserCreate(email="a@b.c", password="asd", slurm_user="  ")
+        UserCreate(email="a@b.c", password="asd", slurm_user="  ")
+    # With valid slurm_accounts attribute
+    u = UserCreate(email="a@b.c", password="asd", slurm_accounts=["a", "b"])
+    assert u.slurm_accounts == ["a", "b"]
+    u = UserCreate(
+        email="a@b.c", password="asd", slurm_accounts=[1, "a", True]
+    )
+    assert u.slurm_accounts == ["1", "a", "True"]
```
tcompa, line 357: I'd rather not have this coerce-to-string feature in-place, but Pydantic v1 doesn't offer a global configuration for strict mode (ref https://docs.pydantic.dev/1.10/usage/validation_decorator/#coercion-and-strictness). However I think that's what `StrictStr` does, right?

```python
from pydantic import BaseModel
from pydantic import StrictStr


class Model(BaseModel):
    x: str


class StrictModel(BaseModel):
    x: StrictStr


print(Model(x=1))
# x='1'

print(StrictModel(x=1))
# Traceback (most recent call last):
#   File "/tmp/x.py", line 14, in <module>
#     print(StrictModel(x=1))
#   File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
# pydantic.error_wrappers.ValidationError: 1 validation error for StrictModel
# x
#   str type expected (type=type_error.str)
```


(ref #1066)

tests/test_unit_schemas.py
```diff
@@ -341,13 +341,27 @@ def test_user_create():
...
+    with pytest.raises(ValidationError):
+        UserCreate(
+            email="a@b.c",
+            password="asd",
+            slurm_accounts=[1, "a", {"NOT": "VALID"}],
```
tcompa, line 362: Let's remove the `1` from this list, to make the failure reason more transparent (and also because of the usual ambiguity related to Pydantic (non-)strict mode).

fractal_server/app/routes/api/v1/project.py
```diff
@@ -315,6 +315,19 @@ async def apply_workflow(
...
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail=(
+                    f"SLURM account '{apply_workflow.slurm_account}' is not "
+                    "among those available to the current user"
+                ),
+            )
+    else:
+        if len(user.slurm_accounts) > 0:
+            apply_workflow.slurm_account = user.slurm_accounts[0]
+
```
tcompa, line 330: What happens when `apply_workflow.slurm_account` is unset? Does it take a `None` value? I think so, but let's check.

fractal_server/app/schemas/user.py
```diff
@@ -54,6 +57,12 @@ class UserUpdate(schemas.BaseUserUpdate):
         val_absolute_path("cache_dir")
     )
 
+    @validator("slurm_accounts")
```
tcompa, line 60: Refactor into a re-usable validator

fractal_server/main.py
```diff
@@ -88,6 +88,7 @@ async def _create_first_user(
     password: str,
     is_superuser: bool = False,
     slurm_user: Optional[str] = None,
+    slurm_accounts: Optional[list[str]] = None,
```
tcompa, line 91: Check that this function is called only once (and that is in this file), and remove all unused arguments: `slurm_user`, `slurm_accounts`, `cache_dir`.

tests/test_security.py
```diff
@@ -45,10 +47,14 @@ async def test_register_user(registered_client, registered_superuser_client):
     res = await registered_superuser_client.post(
         f"{PREFIX}/register/", json=payload_register
     )
-    debug(res.status_code)
-    debug(res.json())
-    assert res.json()["email"] == EMAIL
+    assert res.status_code == 422
```
tcompa, line 50: Maybe move to test_unit_schemas?

fractal_server/app/schemas/applyworkflow.py
```diff
@@ -63,6 +63,7 @@ class ApplyWorkflowCreate(_ApplyWorkflowBase):
 
     first_task_index: Optional[int] = None
     last_task_index: Optional[int] = None
+    slurm_account: Optional[str] = None
```
tcompa, line 66: Let's re-check the `StrictStr` option

fractal_server/app/schemas/user.py
```diff
@@ -44,6 +46,7 @@ class UserUpdate(schemas.BaseUserUpdate):
     slurm_user: Optional[str]
     cache_dir: Optional[str]
     username: Optional[str]
+    slurm_accounts: Optional[list[str]]
```
tcompa, line 49: Let's re-check the `StrictStr` option

fractal_server/app/schemas/user.py
```diff
@@ -93,8 +109,15 @@ class UserCreate(schemas.BaseUserCreate):
     slurm_user: Optional[str]
     cache_dir: Optional[str]
     username: Optional[str]
+    slurm_accounts: list[str] = Field(default_factory=list)
```
tcompa, line 112: Let's re-check the `StrictStr` option

tests/test_security.py
```diff
@@ -110,11 +117,44 @@ async def test_patch_current_user_cache_dir(registered_client):
...
     assert res.json()["cache_dir"] == "/tmp"
+    assert res.json()["slurm_accounts"] == NEW_SLURM_ACCOUNTS
+
+    # slurm_accounts, if present, must be a valid list without repetitions
+    res = await registered_client.patch(
+        f"{PREFIX}/current-user/",
+        json={"slurm_accounts": [42, "Foo", False]},
+    )
+    assert res.status_code == 200
+    assert res.json()["slurm_accounts"] != [42, "Foo", False]
```
tcompa, line 136: Maybe part of this goes to test_unit_schemas?

tests/test_unit_schemas.py
```diff
@@ -341,13 +341,36 @@ def test_user_create():
     u = UserCreate(email="a@b.c", password="asd")
     debug(u)
     assert u.slurm_user is None
+    assert u.slurm_accounts == []
```
tcompa, line 344: Add to this file also the test for `ApplyWorkflow*` schemas

fractal_server/app/routes/api/v1/project.py
tcompa: OK, thanks.

fractal_server/app/schemas/_validators.py
```diff
@@ -63,3 +63,13 @@ def val(string: str):
         return s
 
     return val
+
+
+def val_unique_list(attribute: str):
+    def val(must_be_unique: list):
+        if must_be_unique is not None:
+            if len(set(must_be_unique)) != len(must_be_unique):
+                raise ValueError("`slurm_accounts` list has repetitions")
```
tcompa, line 72: If the error message includes "slurm_accounts", then let's turn this into a standard validator (included in the Pydantic models, rather than re-usable). Or let's change the name to something like `unique_slurm_accounts_list`.

PR #1081, opened 2023-12-13
Add `project` relationship attribute to Dataset(Read) and Workflow(Read)

### LINKED ISSUE #1063: Include `{dataset,workflow}.project` relationship in `{Dataset,Workflow}` model and `{Dataset,Workflow}Read` schema
This would simplify some logic in fractal-web, and sometimes reduce the number of API calls (ref https://github.com/fractal-analytics-platform/fractal-web/issues/359).
Note that a project comes with very little content (three scalar attributes: ID, name and read-only flag), so that this will not significantly increase the response body size. Still: let's make sure that the `ApplyWorkflowRead.project` attribute is of type `ProjectRead`, so that we don't include other data in the response (like `user_list`, `dataset_list`, `workflow_list` and `job_list`).

ychiucco:
closes #1063 

## Checklist before merging

### DIFF
--- a/fractal_server/app/models/dataset.py
+++ b/fractal_server/app/models/dataset.py
@@ -39,7 +39,8 @@ class Dataset(_DatasetBase, SQLModel, table=True):
     id: Optional[int] = Field(default=None, primary_key=True)
     project_id: int = Field(foreign_key="project.id")
     project: "Project" = Relationship(  # noqa: F821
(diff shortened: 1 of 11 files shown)


ychiucco:
I close this pull request and replace it with https://github.com/fractal-analytics-platform/fractal-server/pull/1082

PR #1082, opened 2023-12-14
Add `project` relationship to `Dataset` and `Workflow`

### LINKED ISSUE #1063: Include `{dataset,workflow}.project` relationship in `{Dataset,Workflow}` model and `{Dataset,Workflow}Read` schema

ychiucco:
closes #1063 
replaces #1081 

ref https://github.com/fractal-analytics-platform/fractal-web/issues/359

## Checklist before merging


fractal_server/app/schemas/dataset.py
```diff
@@ -144,6 +144,7 @@ class DatasetRead(_DatasetBase):
     id: int
     resource_list: list[ResourceRead]
     project_id: int
+    project: Optional[ProjectRead]
```
tcompa, line 147: Make non-optional

fractal_server/app/schemas/workflow.py
```diff
@@ -128,6 +128,7 @@ class WorkflowRead(_WorkflowBase):
 
     id: int
     project_id: int
+    project: Optional[ProjectRead]
```
tcompa, line 131: Make non-optional

fractal_server/app/models/project.py
```diff
@@ -24,13 +24,15 @@ class Project(_ProjectBase, SQLModel, table=True):
     )
 
     dataset_list: list[Dataset] = Relationship(
+        back_populates="project",
```
tcompa, line 27: TBD whether this is needed

tcompa:
For the record, this test is passing:
```python
async def test_load_relationships(
    db, project_factory, MockCurrentUser, task_factory
):
    async with MockCurrentUser(persist=True) as user:
        project = await project_factory(user, id=97)
        wf = Workflow(name="wf", project_id=project.id, task_list=[], id=99)
        ds = Dataset(name="ds", project_id=project.id, id=95)
        db.add(wf)
        db.add(ds)
        await db.commit()
        await db.close()

        ###

        stm = (
            select(Project, Workflow, Dataset)
            .filter(Project.id == 97)
            .filter(Workflow.project_id == Project.id)
            .filter(Dataset.project_id == Project.id)
            )
        proj = (await db.scalars(stm)).one()
        debug(type(proj))
       
        for wf in proj.workflow_list:
            debug(WorkflowRead(**wf.__dict__))
        for ds in proj.dataset_list:
            debug(DatasetRead(**ds.__dict__))

        await db.close()

        ###

        proj = await db.get(Project, 97)

        for wf in proj.workflow_list:
            with pytest.raises(ValidationError):
                debug(WorkflowRead(**wf.__dict__))
        for ds in proj.dataset_list:
            with pytest.raises(ValidationError):
                debug(DatasetRead(**ds.__dict__))
```

fractal_server/app/routes/api/v1/dataset.py
```diff
@@ -77,6 +77,8 @@ async def read_dataset_list(
     project = await _get_project_check_owner(
         project_id=project_id, user_id=user.id, db=db
     )
+    for ds in project.dataset_list:
```
tcompa, line 80: Could we avoid the `for` loop?

fractal_server/app/routes/api/v1/workflow.py
```diff
@@ -58,6 +58,9 @@ async def get_workflow_list(
     project = await _get_project_check_owner(
         project_id=project_id, user_id=user.id, db=db
     )
+    for wf in project.workflow_list:
```
tcompa, line 61: Could we avoid the `for` loop?

fractal_server/app/models/project.py
tcompa: Not needed - see test introduced with [07955b1](https://github.com/fractal-analytics-platform/fractal-server/pull/1082/commits/07955b11aefa0e80890340b569e403cdcf7f3d0b).

tcompa:
Regarding the `db.refresh` loop, a first test suggest we should avoid it (and then we can avoid the `dataset_list`):
```python
@router.get(
    "/project/{project_id}/dataset/",
    response_model=list[DatasetRead],
)
async def read_dataset_list(
    project_id: int,
    user: User = Depends(current_active_user),
    db: AsyncSession = Depends(get_db),
) -> Optional[list[DatasetRead]]:
    """
    Get dataset list for given project
    """
    project = await _get_project_check_owner(
        project_id=project_id, user_id=user.id, db=db
    )
    
    # Option 1: select/where
    # Times for endpoint response (local, with sqlite)
    # 10 datasets   -> 0.015 s
    # 100 datasets  -> 0.026 s
    # 1000 datasets -> 0.220 s
    stm = select(Dataset).where(Dataset.project_id == project.id)
    dataset_list = (await db.execute(stm)).scalars().all()

    # Option 2: db.refresh loop
    # Times for endpoint response (local, with sqlite)
    # 10 datasets   -> 0.032 s
    # 100 datasets  -> 0.215 s
    # 1000 datasets -> 2.099 s
    for ds in project.dataset_list:
        await db.refresh(ds)
        
    return project.dataset_list
```

Either option 1 or option 2 above make the test pass, but when used to benchmark the endpoint timing the answer is clearly in favor of select/where - rather than refreshing all elements of the relationship.

tcompa:
A more precise benchmark of the sqlalchemy options is very clearly supporting the behavior described in the previous comment - that is, let's avoid the `db.refresh` loop.

```
$ cat .fractal_server.env 
DB_ENGINE=sqlite
SQLITE_PATH=test.db
```

Benchmark script:
```python
from devtools import debug
import time
import asyncio
from sqlalchemy import select

from fractal_server.app.models import Dataset
from fractal_server.app.models import Project
from fractal_server.app.models import Resource

from fractal_server.app.db import DB
from fractal_server.app.db import get_db
from fractal_server.app.models import SQLModel


async def setup(db, N: int) -> int:
    p = Project(name="proj")
    db.add(p)
    for i in range(N):
        db.add(Dataset(name=str(i), project=p))
    await db.commit()
    await db.refresh(p)
    return p.id


async def teardown(db) -> None:
    res = await db.execute(select(Project))
    for p in res.scalars().all():
        await db.delete(p)
    await db.commit()


async def run_option_1(db, project_id: int) -> float:
    project = await db.get(Project, project_id)
    t_start = time.perf_counter()
    stm = select(Dataset).where(Dataset.project_id == project.id)
    dataset_list = (await db.execute(stm)).scalars().all()
    t_end = time.perf_counter()
    return t_end - t_start


async def run_option_2(db, project_id: int) -> float:
    project = await db.get(Project, project_id)
    t_start = time.perf_counter()
    for ds in project.dataset_list:
        await db.refresh(ds)
    t_end = time.perf_counter()
    return t_end - t_start

async def main():
    DB.set_db()
    engine = DB.engine_sync()
    engine_async = DB.engine_async()
    metadata = SQLModel.metadata
    metadata.create_all(engine)

    async for db in get_db():
        for N in [10, 20, 50, 100, 200, 500, 1000, 2000]:
            project_id = await setup(db, N)
            t1 = await run_option_1(db, project_id)
            t2 = await run_option_2(db, project_id)
            await teardown(db)
            print(f"{N:4d} {t1:.4f} {t2:.4f}")

    metadata.drop_all(engine)
    engine.dispose()
    await engine_async.dispose()


if __name__ == "__main__":
    asyncio.run(main())
```

Result:
```
# Column 1: number of datasets
# Column 2: timing for the select/where option
# Column 3: timing for the db.refresh loop option

  10 0.0097 0.0212
  20 0.0049 0.0386
  50 0.0053 0.0918
 100 0.0059 0.1835
 200 0.0073 0.3824
 500 0.0160 0.9051
1000 0.0221 1.7998
2000 0.0928 3.5820
```

fractal_server/app/routes/api/v1/workflow.py
tcompa: Fixed, as of https://github.com/fractal-analytics-platform/fractal-server/pull/1082/commits/f31b72c6cd95a2d072a7dba9851da4c451e70bbb.

fractal_server/app/routes/api/v1/dataset.py
tcompa: Now fixed

tcompa:
Note that something is probably still missing in the `GET /api/v1/dataset/` endpoint (and perhaps the same holds for workflows):


```
INFO:     127.0.0.1:33808 - "GET /api/v1/dataset/ HTTP/1.1" 500 Internal Server Error
...
    raise ResponseValidationError(
fastapi.exceptions.ResponseValidationError: 2 validation errors:
  {'loc': ('response', 0, 'project'), 'msg': 'field required', 'type': 'value_error.missing'}
  {'loc': ('response', 1, 'project'), 'msg': 'field required', 'type': 'value_error.missing'}
```

The error appears in a real-life usage of Fractal, e.g. through fractal-containers or by running server+web and demos example 01.

tcompa:
fractal-containers CI now passes: https://github.com/fractal-analytics-platform/fractal-containers/actions/runs/7219444094/job/19670474626?pr=35

PR #1099, opened 2023-12-19
Require verified users for task-collection and workflow submission

### LINKED ISSUE #959: Handle OAuth-self-authenticated users
cc @ychiucco

When the `associate_by_email` action (within a OAuth authentication) fails, a new user is created in the DB. This user doesn't have any of the custom Fractal attributes (e.g. `slurm_user`, or the cache directory), and then they would not be able to run a workflow on a SLURM cluster.

Apart from that, should we be more strict in what this user can do with Fractal? The current freedom we have has to do with three user attributes:
* `is_superuser`: currently required for a few specific endpoints
* `is_active`: currently required for most (all?) endpoints
* `is_verified`: currently not used, as far as I remember

If we set the self-authenticated user as non-active, they will have essentially no access to any endpoint. In this case, fractal-web should simply show a big warning page that says something on the line of "Please go and ask the admin to activate your account'.

This is a first proposal, and we can then iterate on it.

tcompa:
Close #959 

## Checklist before merging
- [x] I added specific tests of the verified/non-verified behavior for the relevant endpoint(s).


tests/api/test_admin_api.py
```diff
@@ -20,20 +20,23 @@ async def test_unauthorized_to_admin(client, MockCurrentUser):
     async with MockCurrentUser(
         persist=True, user_kwargs={"is_superuser": False}
     ):
+
         res = await client.get(f"{PREFIX}/project/")
-        assert res.status_code == 403
+        assert res.status_code == 401
```
tcompa, line 25: 403 was our own error status, while fastapi-users uses 401 here

ychiucco (approved):

ychiucco (approved):

PR #1101, opened 2023-12-20
Update testing databases

tcompa:
Close #1073

## Checklist before merging
- [x] I added an appropriate entry to `CHANGELOG.md`
- [x] Use appropriate DB in migrations github action
- [x] Run fix-db script from hard-coded path, for the moment, waiting for #1094

PR #1102, opened 2023-12-20
Add `Project.timestamp_created` column

### LINKED ISSUE #1068: Add `project.timestamp_created` column
The main reason for this change is to be able to sort the project list of a user (that is, `UserOAuth.project_list`) by this new attribute.
Refs:
* https://github.com/fractal-analytics-platform/fractal-server/issues/1062
* https://github.com/fractal-analytics-platform/fractal-web/issues/339

The new attributed should be filled upon project creation (in the corresponding `POST` endpoint), and it should not be editable.

---

This approach won't work by default for projects that already exist, and we will need to prepare the corresponding fix-db script (similar to https://github.com/fractal-analytics-platform/fractal-server/pull/1058).

To be defined: how to assign timestamps to existing projects without knowing their creation time?
Some options:
1. Create fake timestamps which are sorted as the project IDs. This will work reasonably well if there have not been too many project deletions, but it may lead to inconsistencies if some projects were deleted and then their ID were re-assigned.
2. Create fake timestamps based on the jobs in the projects (that is, take the earliest job submission time and use it as a project creation time). This would be more accurate, but we it will be inconsistent for projects which have no jobs (and in that case we have to set a default).
3. A mix of the previous two points -> I tried to come up with a reasonable one, but it seems too complex IMO.

My preference would be to go with 2. All projects with jobs would come up sorted correctly, and all projects with no jobs (which are likely the least interesting ones) would be assigned a fake timestamp like 2023-06-01-00:00:00 and end up in the end of the list.

### LINKED ISSUE #1108: Remove `ApplyWorkflowRead.sanitised_dict`, if obsolete

ychiucco:
closes #1068 
closes #1108 

## Checklist before merging
- [x] I added a fix-db script
- [x] Enrich the fix-db script with extensive schema-based validation (e.g. validate the whole job with `ApplyWorkflowRead`) - ref https://github.com/fractal-analytics-platform/fractal-server/issues/1106. This is useful because it makes errors appear at the fix-db-testing phase (that is, in `migrations.yml` as of #1101), rather than postponing it to when the app is running and some API response cannot be validated.


tcompa:
Let's use `sanitised_dict` when dumping a project, within the apply-workflow endpoint.

fractal_server/utils.py
```diff
@@ -30,6 +30,14 @@ def get_timestamp() -> datetime:
     return datetime.now(tz=timezone.utc)
 
 
+def serialize_timestamp(timestamp: datetime) -> str:
```
tcompa, line 33: Reminder: we still need to review this function

scripts/fix_db/current.py
```diff
@@ -33,7 +35,10 @@
                     f"[Job {job.id:4d}] project_id=None, use dummy data"
                 )
                 project_dump = dict(
-                    id=-1, name="__UNDEFINED__", read_only=True
+                    id=-1,
```
tcompa, line 38: Reminder: this code branch is not currently tested within `migrations.yml`.
Let's modify the test DB.

fractal_server/app/routes/admin.py
```diff
@@ -39,6 +39,8 @@
 async def view_project(
     id: Optional[int] = None,
     user_id: Optional[int] = None,
+    timestamp_created_min: Optional[datetime] = None,
```
tcompa, line 42: We either test this or remove it for now

scripts/fix_db/current.py
ychiucco: Opened its own issue #1113

ychiucco:
I can run `test_project_api` without problems with SQLite v3.39.5

PR #1121, opened 2023-12-22
Set `end_timestamp` when patching Job status

### LINKED ISSUE #1115: Patch-job endpoint should also set the end timestamp
See https://github.com/fractal-analytics-platform/fractal-server/blob/main/fractal_server/app/routes/admin.py.

After setting the status to "failed", we should also set the end timestamp to `datetime.datetime.now(tz=datetime.timezone.utc)`.

ychiucco:
closes #1115 

## Checklist before merging


fractal_server/app/routes/admin.py
```diff
@@ -252,6 +253,7 @@ async def update_job(
         )
 
     setattr(job, "status", job_update.status)
+    setattr(job, "end_timestamp", datetime.now(tz=timezone.utc))
```
tcompa, line 256: We have a function for getting the current timestamp.

tcompa (approved):
I added a minor comment (let's use the existing function for timestamp). Other than that, all good!

PR #1122, opened 2023-12-22
Fix Task-collection endpoint error message

### LINKED ISSUE #1117: Fix task-collection endpoint error message
By setting the package name to `asd.whl`, in the request body, we can get this errro message:

```
"detail": "Invalid package or manifest. Original error: WARNING: Requirement 'asd.whl' looks like a filename, but the file does not exist\nERROR: asd.whl is not a valid wheel filename.\n\n[notice] A new release of pip is available: 23.2.1 -> 23.3.2\n[notice] To update, run: pip install --upgrade pip\n"
```
where the `[notice]` pip-related part should not be here.

ychiucco:
closes #1117 

## Checklist before merging


tcompa:
I think this solution is not sufficiently robust. It relies on at least two assumptions which are not always valid:
1. (trivial) We assume that the wording for "notice" is guaranteed to remain stable in the future (ref https://github.com/pypa/pip/blob/main/src/pip/_internal/self_outdated_check.py#L138).
2. (more relevant) We assume that the warning will _always_ follow the same format, with the relevant part first and the notice part last. If this assumption breaks, we suddenly would loose information about the actual error.

Maybe these assumptions are correct, but we should first verify it.

I suggest that we take a second look, and first of all identify where is the "notice" message coming from. That is, which function raises the error? `download_package`, `inspect_package`, `check`, another one? Once we identify the error source, let's look more carefully at why the "notice" part is there (that is, is it coming from stderr, from stdout, or from somewhere else?). 
With this understanding we may come up with a better solution - or maybe just decide that it's not worth investing in this issue ;)

tcompa:
Closing for now

PR #1136, opened 2024-01-09
Update `_create_first_user` internal function

### LINKED ISSUE #1086: Update `_create_first_user` internal function
If the `is_superuser` argument is `False`, it's probably better to avoid the check of whether a superuser already exists.

~~Or, otherwise, let's fully transform this function into `_create_first_superuser` and simplify its call signature by removing `is_superuser`.~~

ychiucco:
closes #1086 

## Checklist before merging


tcompa:
Inspired by the coverage report, shall we also add a unit test for this function?

I'd say we can move the function to a different subpackage (maybe `app/security/__init__.py`? or `app/security/create_user.py`?) and then we could have `tests/test_unit_security.py`.

Note that `tests/tests_security.py` looks like an API-only module, thus I'd rather move it to `tests/api/test_auth_api.py`.

tcompa:
Are all these FIXMEs in the test code needed for some reason? If so, let's use another keyword. If not, let's drop them.

tests/test_unit_security.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # UserAlreadyExists
+    await _create_first_user(email="test2@fractal.com", password="xxxx")
+    assert await count_users(db) == 2
+    # FIXME test line 322:
+    # logger.warning(f"User {email} already exists")
+
+    await _create_first_user(
+        email="test3@fractal.com", password="xxxx", is_superuser=True
+    )
```
tcompa, line 32: We can also assert that this returns `None`, right?

tests/test_unit_security.py
```diff
@@ -0,0 +1,50 @@
...
+    assert await count_users(db) == 3
+    # can't create more than one superuser
+    await _create_first_user(
```
tcompa, line 35: We can also assert that this returns None, right?

tests/test_unit_security.py
```diff
@@ -0,0 +1,50 @@
...
+    await _create_first_user(email="test1@fractal.com", password="xxxx")
+    assert await count_users(db) == 1
+
+    await _create_first_user(
+        email="test2@fractal.com", password="xxxx", username="test2"
+    )
+    assert await count_users(db) == 2
+
+    # UserAlreadyExists
+    await _create_first_user(email="test2@fractal.com", password="xxxx")
```
tcompa, line 25: To test logs, one should use the `caplog` fixture (see e.g. https://docs.pytest.org/en/7.1.x/how-to/logging.html).

tcompa: See also https://github.com/search?q=repo%3Afractal-analytics-platform%2Ffractal-server%20caplog&type=code

tests/test_unit_security.py
ychiucco: The return value of `_create_first_user` is always None

tcompa: That's right, my bad

tcompa (approved):

PR #1137, opened 2024-01-09
WorkflowTask foreign keys not nullables

### LINKED ISSUE #1095: Make `WorfklowTask` foreign keys not nullable
Currently the database accepts `WorfklowTask.workflow_id` and `WorfklowTask.task_id` equal to `null`, but Fractal does not allow the creation of "orphan" WorkflowTasks.

ychiucco:
closes #1095 

## Checklist before merging


fractal_server/app/models/workflow.py
```diff
@@ -49,8 +49,10 @@ class Config:
 
     id: Optional[int] = Field(default=None, primary_key=True)
 
-    workflow_id: Optional[int] = Field(foreign_key="workflow.id")
-    task_id: Optional[int] = Field(foreign_key="task.id")
+    workflow_id: Optional[int] = Field(
```
tcompa, line 52: Why using `Optional` in the type hint and then `nullable=False`?
I think we should at least remove `Optional`, and probably also remove `nullable=False`.

fractal_server/app/models/workflow.py
```diff
@@ -49,8 +49,10 @@ class Config:
...
+        foreign_key="workflow.id", nullable=False
+    )
+    task_id: Optional[int] = Field(foreign_key="task.id", nullable=False)
```
tcompa, line 55: Same question as for `workflow_id`: why `Optional` and then `nullable=False`?

fractal_server/app/models/workflow.py
```diff
@@ -139,12 +141,19 @@ async def insert_task(
             db: TBD
             commit: TBD
         """
+        if self.id is None:
+            raise ValueError(
+                "Cannot use `insert_task` for a non-committed Workflow"
```
tcompa, line 146: I suggest we go with an even simpler message (independent on whether the workflow was committed or maybe something else went wrong).
Example: "Cannot use `Workflow.insert_task` when `Workflow.id=None`"

fractal_server/app/models/workflow.py
```diff
@@ -161,7 +170,12 @@ async def insert_task(
             wt_meta = None
 
         # Create DB entry
-        wf_task = WorkflowTask(task_id=task_id, args=actual_args, meta=wt_meta)
+        wf_task = WorkflowTask(
+            workflow_id=self.id,
```
tcompa, line 174: This line (`workflow_id=self.id`) looks perfectly correct, but I'm confused: how could the previous version work? Where did we establish this foreign-key?

mfranzon: The suspicion is that since it is `Optional` it is often null... can we verify it in a db?

tcompa: On a realistic DB (and from within psql):
```
# select id,workflow_id,task_id from workflowtask;
  id  | workflow_id | task_id 
------+-------------+---------
...
(678 rows)

# select id,workflow_id,task_id from workflowtask where workflow_id is null;
 id | workflow_id | task_id 
----+-------------+---------
(0 rows)

# select id,workflow_id,task_id from workflowtask where task_id is null;
 id | workflow_id | task_id 
----+-------------+---------
(0 rows)
```

tcompa: > This line (`workflow_id=self.id`) looks perfectly correct, but I'm confused: how could the previous version work? Where did we establish this foreign-key?

I'll answer myself: it was working because of
```python
class Workflow(_WorkflowBase, SQLModel, table=True):
    ...
    task_list: list[WorkflowTask] = Relationship(...)
    ...
    async def insert_task(...):
        ...
        self.task_list.insert(order, wf_task)
```
which also updates the (list of) relationship.

tcompa (approved):

Issue #1094, opened 2023-12-19
Review location of fix-db scripts


tcompa:
After the 1.3->1.4.0 version update, we ran some fix-db scripts from the `scripts/database_updates/from_1.3_to_1.4` folder.
As of #1070 (and in future PR related to #1068), we'll have one or two more such scripts.

Having them in an arbitrary `scripts` folder is not great for usability, not for reliability. I propose that, together with #1073, we also test the scripts from the migrations github action.

A plan could be:
1. Move the scripts within an appropriate subfolder of the package (e.g. `fractal_server/fix_db_scripts`).
2. Turn the scripts into functions.
3. Make the functions version-aware. This could be done in multiple ways, and an example is that a given function would first check that the version matches a certain regex (e.g the script introduced in #1070 should be only applied if version is `>=1.4.2`, and we may even want to require that it is strictly equal to 1.4.2 - up to alpha prereleases).
4. Expose the appropriate function via the `fractalctl` CLI. An example procedure would be:
    * `pip install fractal-server==1.4.2`
    * `fractalctl set-db`
    * `fractalctl fix-db 1.4.2` -> this runs the appropriate function
    * `fractalctl start`
5. Add the same procedure (`set-db` and then `fix-db` to `migrations.yml` test). This obviously requires keeping an up-to-date test DB, as in #1073).

I think that with this versioning approach we could even avoid keeping track of old scripts, and only keep the one needed for the last transition (to be verified). If possible, that'd be preferable, so that we don't have obsolete/untested code in the package and repo.

tcompa:
Here is an example of why this issue matters:

The current script has a tiny bug:
```diff
diff --git a/scripts/database_updates/from_1.4.1_to_1.4.2/fix_db_add_project_dump.py b/scripts/database_updates/from_1.4.1_to_1.4.2/fix_db_add_project_dump.py
index 59a458d8..5b363456 100644
--- a/scripts/database_updates/from_1.4.1_to_1.4.2/fix_db_add_project_dump.py
+++ b/scripts/database_updates/from_1.4.1_to_1.4.2/fix_db_add_project_dump.py
@@ -39,7 +39,7 @@ with next(get_sync_db()) as db:
             logging.warning(
                 f"Handling job with ID {row.id:4d}, setting {project_dump=}"
             )
-            ProjectDump(project_dump)
+            ProjectDump(**project_dump)
             row.project_dump = project_dump
             db.add(row)
             db.commit()
```
[...]

PR #1152, opened 2024-01-11
Add `timestamp_created` to `Workflow` and `Dataset`

### LINKED ISSUE #1118: Add `timestamp_created` column for workflow and dataset tables
This should follow the same approach as for the `project` table, ref
* https://github.com/fractal-analytics-platform/fractal-server/issues/1068
* https://github.com/fractal-analytics-platform/fractal-server/pull/1102

In the fix-db script, we can simply defer the fix to the `project.timestamp_created` part (by doing the project fix first, and the workflow and dataset fixes next). For instance if we find the 1-1-1-0-0-0-0 timestamp for a given `dataset.timestamp_created`, then we read the corresponding project (note that `dataset.project_id` is not nullable) and we use `project.timestamp_created`. The same holds for `workflow.timestamp_created`.

Note that in principle we'd also have to fix the dump Pydantic schemas with the new timestamps. Or maybe we can decide that timestamps are not really important in dump objects, since the `applyworkflow` object already has its own start/end timestamps.

ychiucco:
closes #1118 

## Checklist before merging


.github/workflows/migrations.yml
```diff
@@ -15,7 +15,7 @@ on:
 
 
 env:
-  TARGET_VERSION: 1.4.1
+  TARGET_VERSION: 1.4.2
```
tcompa, line 18: It's maybe useful to add a comment here, given our current internal discussions.. Something that says that the DB was created with this version of fractal-server, and is now used while upgrading to the current version.

CHANGELOG.md
```diff
@@ -1,10 +1,21 @@
 **Note**: Numbers like (\#123) point to closed Pull Requests on the fractal-server repository.
 
+# 1.4.3
+
```
tcompa, line 4: Let's add here the note about running the fix-db script, as for 1.4.2.

CHANGELOG.md
```diff
@@ -1,10 +1,21 @@
...
+* API:
+    * Add `timestamp_created` to `WorkflowRead`, `WorkflowDump`, `DatasetRead` and `DatasetDump` (\#1152).
+* Database:
+    * Add `Workflow.timestamp_created` and `Dataset.timestamp_created` columns (\#1152).
+* Testing:
+    * Update `clean_db_fractal_1.4.1.sql` to `clean_db_fractal_1.4.2.sql`, and change `migrations.yml` target version (\#1152).
+    * Start a new `current.py` fix-db script (\#1152).
```
tcompa, line 11: Rather than "Testing", this item better fits within "Database"

fractal_server/app/routes/api/v1/project.py
```diff
@@ -395,21 +395,28 @@ async def apply_workflow(
...
+            timestamp_created=str(output_dataset.timestamp_created),
             resource_list=[
                 resource.model_dump()
                 for resource in output_dataset.resource_list
             ],
         ),
         workflow_dump=dict(
-            workflow.model_dump(exclude={"task_list"}),
+            workflow.model_dump(exclude={"task_list", "timestamp_created"}),
+            timestamp_created=str(workflow.timestamp_created),
```
tcompa, line 419: This way of serializing `timestamp_created` is fine, but it doesn't match with what we do for `project_dump` a few lines below. Can we make them homogeneous?
The current approach (exclude, and then use `str`) looks simpler than `json.loads(project.json(...))`, thus I'm OK with having it also for `project_dump` - unless I'm missing some other issues.

scripts/fix_db/current.py
```diff
@@ -11,92 +6,212 @@
...
-            project.timestamp_created = timestamp_created
-            db.add(project)
+            dataset.timestamp_created = new_timestamp
+            db.add(dataset)
             db.commit()
-            db.refresh(project)
-            db.expunge(project)
-            ProjectRead(**project.model_dump())
+            db.refresh(dataset)
+            db.expunge(dataset)
```
tcompa, line 89: Why do we need to refresh and expunge?

scripts/fix_db/current.py
```diff
@@ -11,92 +6,212 @@
...
             )
-            jobs = db.execute(stm).scalars().all()
-            if len(jobs) == 0:
-                logging.warning(
-                    f"[Project {project.id:4d}] No jobs found, skip."
+            workflow.timestamp_created = new_timestamp
+            db.add(workflow)
+            db.commit()
+            db.refresh(workflow)
+            db.expunge(workflow)
```
tcompa, line 55: Why do we need to refresh and expunge?

scripts/fix_db/current.py
```diff
@@ -1,8 +1,3 @@
-"""
-Loop over jobs.
-If the corresponding project still exists, set the project_dump.
-"""
-import json
 import logging
```
tcompa, line 1: General comment on this file: there a bunch of commented lines. Are they needed? If not, let's remove them.

CHANGELOG.md
```diff
@@ -1,14 +1,22 @@
 **Note**: Numbers like (\#123) point to closed Pull Requests on the fractal-server repository.
 
-
 # 1.4.3 (unreleased)
 
+> **WARNINGs**:
+>
+> 1. This update requires running a fix-db script, available at https://raw.githubusercontent.com/fractal-analytics-platform/fractal-server/main/scripts/fix_db/current.py.
```
tcompa, line 7: This URL is risky, cause it does not include the version tag. It's best to use a versioned URL, as we did for 1.4.2.

scripts/fix_db/current.py
ychiucco: Commented lines are intended to temporarly skip this two validations
```python
WorkflowRead(**workflow.model_dump())

DatasetRead(**dataset.model_dump())
```
By introducing them, we get this error in `migrations.yml`
```shell
Traceback (most recent call last):
  File "/home/runner/work/fractal-server/fractal-server/scripts/fix_db/current.py", line 55, in <module>
    WorkflowRead(**workflow.model_dump())
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 2 validation errors for WorkflowRead
task_list
  field required (type=value_error.missing)
project
  field required (type=value_error.missing)
Error: Process completed with exit code 1.
```
it seems like we are not loading the relationships along with the main objects.

CHANGELOG.md
ychiucco: Not having a 1.4.3 release yet, I had thought of leaving `main` until the release is ready

tcompa: You're right, let's wait.

Also, for the record, as of https://github.com/fractal-analytics-platform/fractal-server/issues/1094#issuecomment-1893931516 we will likely move away from `current.py` and write the version name in the Python module itself - so that this risk (downloading or running the wrong script) won't be there any more.

scripts/fix_db/current.py
tcompa: This is a known issue: https://github.com/fractal-analytics-platform/fractal-server/issues/489. It's likely that there are more elegant solutions (and they will possibly change in Pydantic v2), but I think the following logic may already work (to check).

```python
from sqlalchemy import select
from fractal_server.app.db import get_sync_db
from fractal_server.app.models import Workflow
from fractal_server.app.schemas import WorkflowRead


with next(get_sync_db()) as db:
    stm = select(Workflow)
    workflows = db.execute(stm).scalars().all()
    for workflow in workflows:
        assert workflow.task_list is not None
        assert workflow.project is not None

        # Option 1 (fails)
        WorkflowRead(**workflow.model_dump())
        
        # Option 2 (goes through)
        project = workflow.project
        task_list = workflow.task_list
        WorkflowRead(**workflow.model_dump(exclude={"task_list", "project"}),  task_list=task_list, project=project)
```

scripts/fix_db/current.py
```diff
@@ -11,92 +6,217 @@
...
+            dataset.timestamp_created = new_timestamp
+            db.add(dataset)
             db.commit()
-            db.refresh(project)
-            db.expunge(project)
-            ProjectRead(**project.model_dump())
+            db.refresh(dataset)
+            db.expunge(dataset)
+            DatasetRead(
+                **dataset.model_dump(),
```
tcompa, line 93: I suggest we explicitly exclude the attributes which are then included manually, for clarity.

PR #1166, opened 2024-01-15
Change test execution order

### LINKED ISSUE #1100: Change test-execution order (and improve their organization)
As of #1090, we now have a better organization of tests - including at least subfolders `db`, `api` and `backend`.

We could also add:
* ~~`tests/schemas`, with the tests for Pydantic schemas~~ EDIT: this is already in-place
* `tests/unit`, with tests of function which do not belong to any other subfolder

Tests in subfolders are typically faster and more low-level than the rest, and they should run first. This will make the CI fail faster for the large fraction of PRs that only change some specific part of the schemas, models or API.

 A reasonable order could be like
1. unit
2. schemas
3. db
4. api
5. backend (which include SLURM)
6. all the non-subfolder tests

The ordering feature should be available through https://github.com/pytest-dev/pytest-order (I did not try it out yet).

EDIT: see https://github.com/pytest-dev/pytest-order/pull/52 or https://github.com/pytest-dev/pytest-order/discussions/69

ychiucco:
closes #1100 

## Checklist before merging


tcompa:
I guess that pytest-order was not working.. to bad, but thanks for trying! Thus I agree that it's fine to fix this issue as part of the `ci.yml` file.

I have a couple of comments:

1)
Is there a reason for using
```
poetry run coverage run -p --concurrency=thread,greenlet,multiprocessing -m pytest -k "tests/api"
```
rather than one of
```
poetry run coverage run -p --concurrency=thread,greenlet,multiprocessing -m pytest tests/api
poetry run coverage run -p --concurrency=thread,greenlet,multiprocessing -m pytest tests/api/*.py
```
?

I find the latter two options a bit more transparent, if one doesn't want to go look in the pytest documentation.

2)
I'd rather avoid the fact that tests in `tests/` are not run unless they are part of a subfolder. Also: the `others` subfolder is not really relevant..
If we follow point 1, then we could simply add
```
poetry run coverage run -p --concurrency=thread,greenlet,multiprocessing -m pytest tests/*.py
```
as a last line, and in my understanding we wouldn't need to use the `others` folder at all.

3)
Could you please sort test groups as detailed in the issue? Or was there a reason for changing the order?

tcompa:
Ref https://github.com/fractal-analytics-platform/fractal-server/issues/1100#issuecomment-1895765277

tcompa:
This PR works, but it introduces an important regression in the CI total time:
image

I think we should find a better way, or close this.

tcompa:
What about https://pytest-order.readthedocs.io/en/latest/configuration.html#order-marker-prefix? Would it be an option?

Downside: a lot of decorator. This can be maybe solved with https://stackoverflow.com/a/68573824/19085332?

tcompa (approved):

PR #1178, opened 2024-01-18
1128 create testing db programmatically

### LINKED ISSUE #1128: Create testing db programmatically
We often need a testing database, which we may need to modify in multiple ways, e.g. in order to
- Align with updated database schemas in new versions
- Make all branches of fix-db scripts testable (ref #1113)
- Create different environments for API benchmarks, e.g. two users with few/many DB objects (ref #1089)

We repeated this operation multiple times in previous and current PRs (e.g. #1120 and #1123), typically by acting on the DB (either directly or through fractal-server).
We need to:
1. Make this easily available as a utility in fractal-server, since it's a very frequent operation.
2. Create the DB by working through the API (rather than working directly on the DB), to keep it as close as possible to a real-life case.

### How to proceed

We simply need an appropriate client, and then a set of function/scripts that perform typical flows. Typical examples:
* A function that creates a dummy task;
* A function that creates a user and their project/workflows/datasets and then runs a few successful/failed jobs.
* Functions to delete different kinds of objects, therefore creating null foreign keys for jobs.

As for the client to use: if fractal-client was part of the fractal-server package, that would be the obvious choice. Right now, it'd be too much of an effort (ref https://github.com/fractal-analytics-platform/fractal-client/issues/545), so let's take a shortcut. We can either copy-paste part of https://github.com/fractal-analytics-platform/fractal-client/blob/main/fractal_client/authclient.py, or just rewrite a simple client (e.g. via httpx) that we can use for this goal.

mfranzon:
## Checklist before merging

closes #1128


scripts/populate_db/populate_first_version.py
```diff
@@ -0,0 +1,43 @@
...
+from fractal_server.app.schemas import ResourceCreate
+from fractal_server.app.schemas import TaskCreate
+from fractal_server.app.schemas import WorkflowCreate
+from fractal_server.app.schemas import WorkflowTaskCreate
+
+
+if __name__ == "__main__":
+    client = SimpleHttpClient(base_url=BASE_URL, credentials=CREDENTIALS)
+
+    x = New(client=client)
```
tcompa, line 19: New -> FractalClient

scripts/populate_db/populate_clients.py
```diff
@@ -0,0 +1,154 @@
...
+        res = self.client.make_request(
+            endpoint=f"api/v1/project/{project_id}/workflow/"
+            f"{workflow_id}/wftask/?{task_id=}",
+            method="POST",
+            data=wftask.dict(),
+        )
+        self.detail(res.json())
+
+        return WorkflowTaskRead(**res.json())
+
```
tcompa, line 124: Let's also include  and `add_working_task` (command `echo`) and `add_failing_task` (command `ls`)

scripts/populate_db/populate_clients.py
```diff
@@ -0,0 +1,154 @@
...
+                raise ValueError(f"Unsupported HTTP method: {method}")
+
+            return response
+
+        except requests.RequestException as e:
+            print(f"Request failed: {e}")
+            return None
+
+
+class New:
```
tcompa, line 59: New method: `wait_for_all_jobs`.

if called as superuser, it can use directly `/admin/job/`, with a query param for status.

within a while loop, this can be blocking until all jobs are failed or done

scripts/populate_db/populate_clients.py
```diff
@@ -0,0 +1,154 @@
...
+        res = self.client.make_request(
+            endpoint="api/v1/task/",
+            method="POST",
+            data=task.dict(exclude_none=True),
+        )
+        self.detail(res.json())
+
+        return TaskRead(**res.json())
+
+    def add_job(
```
tcompa, line 136: -> apply_workflow

tcompa:
This is great. Let's have four users:

1. The default superuser `admin@fractal.xy`, only used to create tasks and other users (AKA no project). Tasks should both go from Any to Any, and be one working and one failing
2. `simple_user@example.org` with one project, one dataset (for both I/O), one workflow, one job
3. `power_user@example.org` with one project, one dataset (for both I/O), 10 workflow (5 with working task, 5 with failing task), 20 jobs/workflow
4. `complex_user@example.org`, with twice as many resources as `simple_user` - but we delete the second project at the end, to break some relationships

1 would be a service user
2&3 would form the baseline and actual scenario for benchmarks
4 would be the place to stress-test `migrations.yml`

jluethi:
> 2&3 would form the baseline and actual scenario for benchmarks

If we do benchmarking scenarios, let's also include:
5. A many datasets user:A user with 20 datasets in the same project and 20 workflows, 10 jobs/workflow
6. A many projects user:A user with 25 projects, each with some content (e.g. each with 2 datasets, 1 workflow, 5 jobs)
7. A many jobs user: a user with one project, one dataset (for both I/O), one workflow, 100 jobs

PR #1179, opened 2024-01-18
Remove `RUNNING` from `JobStatusType`

### LINKED ISSUE #1085: Remove `RUNNING` from `JobStatusType`?
In this part:

```
        SUBMITTED:
            The workflow has been applied but not yet scheduled with an
            executor. In this phase, due diligence takes place, such as
            creating working directory, assemblying arguments, etc.
        RUNNING:
            The workflow was scheduled with an executor. Note that it might not
            yet be running within the executor, e.g., jobs could still be
            pending within a SLURM executor.
```

I think it's somewhat misleading to call it RUNNING if it may be scheduled with slurm, but not actually running. And the differentiation on whether it has gone through our due diligence isn't very valuable to a typical user.

_Originally posted by @jluethi in https://github.com/fractal-analytics-platform/fractal-server/issues/1072#issuecomment-1854228452_
            

---

Current situation:
1. Jobs are always created with a "submitted" status.
2. When the API call is over, a background task starts the actual execution of the workflow, and in its pre-execution phase it also sets the status to "running" (which, in the current form, means "some Fractal backend executor is now taking care of executing the job", but not "SLURM is running the job").

I agree that "running" could be confusing (due to overlap with SLURM status), and it does not add very relevant information. In view of future plans of integrating SLURM information somewhere in the API, I agree with removing "running" and sticking with "submitted".

ychiucco:
closes #1085 

## Checklist before merging

### DIFF
--- a/fractal_server/app/routes/api/v1/dataset.py
+++ b/fractal_server/app/routes/api/v1/dataset.py
@@ -374,10 +374,11 @@ async def export_history_as_workflow(
     )
     dataset = output["dataset"]
 
-    # Check whether there exists an active job such that `job.output_dataset_id
-    # == dataset_id`. If at least one such job exists, then this endpoint will
-    # fail. We do not support the use case of exporting a reproducible workflow
-    # when job execution is in progress; this may change in the future.
+    # Check whether there exists a submitted job such that
+    # `job.output_dataset_id==dataset_id`. If at least one such job exists,
+    # then this endpoint will fail.
+    # We do not support the use case of exporting a reproducible workflow when
+    # job execution is in progress; this may change in the future.
     stm = _get_submitted_jobs_statement().where(
         ApplyWorkflow.output_dataset_id == dataset_id
     )
--- a/fractal_server/app/routes/api/v1/dataset.py
+++ b/fractal_server/app/routes/api/v1/dataset.py
@@ -28,9 +28,9 @@ from ....schemas import WorkflowExport
 from ....schemas import WorkflowTaskExport
 from ....security import current_active_user
 from ....security import User
-from ._aux_functions import _get_active_jobs_statement
 from ._aux_functions import _get_dataset_check_owner
 from ._aux_functions import _get_project_check_owner
+from ._aux_functions import _get_submitted_jobs_statement
 from ._aux_functions import _get_workflow_check_owner
 
 
@@ -173,7 +173,7 @@ async def delete_dataset(
 
     # Fail if there exist jobs that are active (that is, pending or running)
     # and in relation with the current dataset.
-    stm = _get_active_jobs_statement().where(
+    stm = _get_submitted_jobs_statement().where(
         or_(
             ApplyWorkflow.input_dataset_id == dataset_id,
             ApplyWorkflow.output_dataset_id == dataset_id,
@@ -378,7 +378,7 @@ async def export_history_as_workflow(
     # == dataset_id`. If at least one such job exists, then this endpoint will
     # fail. We do not support the use case of exporting a reproducible workflow
     # when job execution is in progress; this may change in the future.
-    stm = _get_active_jobs_statement().where(
+    stm = _get_submitted_jobs_statement().where(
         ApplyWorkflow.output_dataset_id == dataset_id
     )
     res = await db.execute(stm)
@@ -449,7 +449,7 @@ async def get_workflowtask_status(
     # Note: see
     # https://sqlmodel.tiangolo.com/tutorial/where/#type-annotations-and-errors
     # regarding the type-ignore in this code block
-    stm = _get_active_jobs_statement().where(
+    stm = _get_submitted_jobs_statement().where(
         ApplyWorkflow.output_dataset_id == dataset_id
     )
     res = await db.execute(stm)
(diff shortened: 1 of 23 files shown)


tests/05_backend/test_backend_slurm.py
```diff
@@ -184,13 +184,13 @@ def wait_and_return():
             fut = executor.submit(wait_and_return)
             debug(fut)
 
-            # Wait until the SLURM job goes from PENDING to RUNNING
+            # Wait until the SLURM job goes from PENDING to SUBMITTED
```
tcompa, line 187: Note that in this module RUNNING corresponds to a SLURM status, and it should not be modified.

fractal_server/app/routes/api/v1/dataset.py
```diff
@@ -173,7 +173,7 @@ async def delete_dataset(
 
     # Fail if there exist jobs that are active (that is, pending or running)
```
tcompa, line 174: Please also fix comment, using "submitted"

fractal_server/app/routes/api/v1/dataset.py
```diff
@@ -378,7 +378,7 @@ async def export_history_as_workflow(
     # == dataset_id`. If at least one such job exists, then this endpoint will
     # fail. We do not support the use case of exporting a reproducible workflow
     # when job execution is in progress; this may change in the future.
-    stm = _get_active_jobs_statement().where(
+    stm = _get_submitted_jobs_statement().where(
```
tcompa, line 381: Please also fix comment above this statement, using "submitted"

fractal_server/app/routes/api/v1/workflow.py
```diff
@@ -194,7 +194,7 @@ async def delete_workflow(
 
     # Fail if there exist jobs that are active (that is, pending or running)
```
tcompa, line 195: Please also fix comment, using "submitted"

fractal_server/app/schemas/applyworkflow.py
```diff
@@ -25,21 +25,14 @@ class JobStatusType(str, Enum):
...
     Attributes:
         SUBMITTED:
-            The workflow has been applied but not yet scheduled with an
-            executor. In this phase, due diligence takes place, such as
-            creating working directory, assemblying arguments, etc.
-        RUNNING:
-            The workflow was scheduled with an executor. Note that it might not
-            yet be running within the executor, e.g., jobs could still be
-            pending within a SLURM executor.
+            TBD
```
tcompa, line 28: We can use something like

The job was created. This does not guarantee that it was also submitted to an executor (e.g. other errors could have prevented this), nor that it is actually running (e.g. SLURM jobs could be still in the queue).

tests/04_api/test_dataset_api.py
```diff
@@ -260,12 +260,6 @@ async def test_delete_dataset_cascade_jobs(
             status=JobStatusType.SUBMITTED,  # this is why ds are not deletable
             **common_args,
         )
-        await job_factory(
-            input_dataset_id=ds_not_deletable_2.id,
-            output_dataset_id=ds_not_deletable_1.id,
```
tcompa, line 265: If we remove this, why is `ds_non_deletable_1` non deletable?

tcompa: Note that there is probably a small bug a few lines below, as we are making the same call twice:
```python
        res = await client.delete(
            f"api/v1/project/{project.id}/dataset/{ds_not_deletable_1.id}/"
        )
        assert res.status_code == 422
        res = await client.delete(
            f"api/v1/project/{project.id}/dataset/{ds_not_deletable_1.id}/"
        )
        assert res.status_code == 422
```

tcompa (approved):

tests/04_api/test_dataset_api.py
ychiucco: > If we remove this, why is `ds_non_deletable_1` non deletable?

because of this `fractal_server/app/routes/api/v1/dataset.py:
```python
# Fail if there exist jobs that are submitted and in relation with the 
# current dataset.
stm = _get_submitted_jobs_statement().where(
    or_(
        ApplyWorkflow.input_dataset_id == dataset_id,
        ApplyWorkflow.output_dataset_id == dataset_id,
    )
)
res = await db.execute(stm)
jobs = res.scalars().all()
if jobs:
    string_ids = str([job.id for job in jobs])[1:-1]
    raise HTTPException(
        detail=(
            f"Cannot delete dataset {dataset.id} because it "
            f"is linked to active job(s) {string_ids}."
        ),
    )
```
and
```python
await job_factory(
    input_dataset_id=ds_not_deletable_1.id,
    output_dataset_id=ds_not_deletable_2.id,
    status=JobStatusType.SUBMITTED,  # this is why ds are not deletable
    **common_args,
)
```
so both `ds_not_deletable_1` and `ds_not_deletable_2` can't be deleted

tcompa: Thanks for clarifying!

(my mistake is that was rather thinking about the case where we prevent two jobs from running with the same _output_ dataset, and in that case the _input_ dataset is not relevant).

PR #1199, opened 2024-01-29
Add rate limit to apply-workflow endpoint

### LINKED ISSUE #1174: To explore: add rate limit to apply-workflow endpoint
Somewhat related to:
* #1140
* https://github.com/fractal-analytics-platform/fractal-web/issues/387

Should we rate-limit calls to the apply-workflow endpoint? Especially if they receive the same payload, meaning they were likely a mistake.

High-level goal: a given user cannot submit the same API request twice in a short time frame (e.g. 3 seconds). The identity of the request can be based e.g. on the combination of user + request body parameters (+ maybe request query parameters).

To do:
- review the current "rate limit" approach, based on timestamps and folder names
- check whether it can be improved, by querying the DB for jobs with the same foreign keys and with a timestamp close to the current one
- or whether it's best to use one of the many external libraries that offer this feature

ychiucco:
closes #1174 

## Checklist before merging


fractal_server/app/routes/api/v1/project.py
```diff
@@ -430,6 +431,35 @@ async def apply_workflow(
...
+        if any(
+            abs(
+                db_job.start_timestamp.replace(tzinfo=timezone.utc)
+                - job.start_timestamp
+            )
+            < timedelta(seconds=1)
+            for db_job in db_jobs
+        ):
+            raise HTTPException(
+                status_code=status.HTTP_409_CONFLICT,
```
tcompa, line 454: I think the appropriate status code would be 429

fractal_server/app/routes/api/v1/project.py
```diff
@@ -430,6 +431,35 @@ async def apply_workflow(
...
+    )
+    res = await db.execute(stm)
+    db_jobs = res.scalars().all()
+    if db_jobs:
+        if any(
+            abs(
+                db_job.start_timestamp.replace(tzinfo=timezone.utc)
+                - job.start_timestamp
+            )
+            < timedelta(seconds=1)
```
tcompa, line 450: Let's move this to an optional config variable. Something like
```python
FRACTAL_API_SUBMIT_MIN_WAIT: int = 2
```

tests/04_api/test_project_apply_api.py
```diff
@@ -631,3 +633,65 @@ async def test_project_apply_slurm_account(
             json=dict(slurm_account="NOT IN THE LIST"),
         )
         assert res.status_code == 422
+
+
+async def test_rate_limit(
```
tcompa, line 638: I'd suggest that only for this test we set `FRACTAL_API_SUBMIT_MIN_WAIT=1`, and we set it to 0 by default in all other tests (through the usual fixture). In this way the other tests remain unchanged, and we don't need to add any `sleep`.

fractal_server/app/routes/api/v1/project.py
```diff
@@ -430,6 +431,37 @@ async def apply_workflow(
...
+    # same database keys (Project, Workflow and Datasets) during the last
+    # `settings.FRACTAL_API_SUBMIT_RATE_LIMIT` seconds.
+    stm = (
+        select(ApplyWorkflow)
+        .where(ApplyWorkflow.project_id == project_id)
+        .where(ApplyWorkflow.workflow_id == workflow_id)
+        .where(ApplyWorkflow.input_dataset_id == input_dataset_id)
+        .where(ApplyWorkflow.output_dataset_id == output_dataset_id)
+        .order_by(ApplyWorkflow.start_timestamp.desc())
+        .limit(1)
```
tcompa, line 446: Should we also include some of the `**apply_workflow.dict()` attributes in this check?

tcompa, line 446: Could you explain how the `order_by` and `limit` options work together? If we are sorting by timestamp, why would we only keep one of the output items?

ychiucco: Of all the jobs with these foreign keys, the only one we need is the most recent one, so taking others would be redundant.

fractal_server/config.py
```diff
@@ -348,6 +348,11 @@ def make_FRACTAL_TASKS_DIR_absolute(cls, v):
     Default values correspond to `vite` defaults.
     """
 
+    FRACTAL_API_SUBMIT_RATE_LIMIT: int = 2
+    """
+    TBD
```
tcompa, line 353: Missing docstring.

PR #1204, opened 2024-01-30
Split tests in two groups

### LINKED ISSUE #1202: Split tests in two groups, to speed GitHub CI
We should try and see how it works to use the same as https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/main/.github/workflows/ci_poetry.yml. For instance we could extract the backend and full-workflow tests into a different "job", and see whether the overhead of installing things twice is compensated by the increased parallelization.

Note that we should always check that coverage is computed correctly - see example above.

ychiucco:
closes #1202 

## Checklist before merging

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
           poetry install --with dev --without docs --no-interaction -E slurm -E gunicorn $DB
 
       - name: Test with pytest
-        run: poetry run coverage run --concurrency=thread,greenlet,multiprocessing -m pytest tests/0[123467]*
+        run: poetry run coverage run --concurrency=thread,greenlet,multiprocessing -m pytest --ignore tests/05_backend --ignore tests/08_full_workflow
 
       - name: Upload coverage data
         uses: actions/upload-artifact@v3
@@ -69,7 +69,7 @@ jobs:
           name: coverage-data
           path: ".coverage*"
 
-  tests_job:
+  tests_runner:
 
     name: "Job - Python ${{ matrix.python-version }} - ${{ matrix.db }}"
     runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -62,7 +123,7 @@ jobs:
           poetry install --with dev --without docs --no-interaction -E slurm -E gunicorn $DB
 
       - name: Test with pytest
-        run: poetry run coverage run --concurrency=thread,greenlet,multiprocessing -m pytest
+        run: poetry run coverage run --concurrency=thread,greenlet,multiprocessing -m pytest tests/0[58]*
```
tcompa, line 126: I'd rather achieve the same split in a more explicit way, if possible.
See https://github.com/fractal-analytics-platform/fractal-tasks-core/blob/main/.github/workflows/ci_poetry.yml#L46C54-L46C74.

In the "core" block, we could use `pytest --ignore X --ignore Y`, and in the "job" block we could use `pytest folder1 folder2`

.github/workflows/ci.yml
```diff
@@ -18,6 +18,67 @@ jobs:
...
+      - name: Test with pytest
+        run: poetry run coverage run --concurrency=thread,greenlet,multiprocessing -m pytest tests/0[123467]*
+
+      - name: Upload coverage data
+        uses: actions/upload-artifact@v3
+        with:
+          name: coverage-data
+          path: ".coverage*"
+
+  tests_job:
```
tcompa, line 72: Better naming: "tests_runner"

tcompa:
This looks great! There is some overhead, of course, but it's not so bad. The total pytest time is essentially unchanged, while the rest of the action takes ~40 seconds.

Here are some examples, for the record.

# Old CI: 8m48s
image

# New CI, core block: 5m15s
image

# New CI, runner block: 4m27s
image

tcompa (approved):

PR #1238, opened 2024-02-20
Remove image list from within `write_args_file`

### LINKED ISSUE #1237: Do not write image list to args.json files
Related to #474.

TLDR: let's remove the `args["metadata"]["image"]` key-value pair before writing args.json files to disk.

---

Here is a real-life example (job ID 1029):

- The workflow has nine tasks (from 0 to 8).
- The dataset has 357 wells and 26 cycles, corresponding to 9282 images (these values are read for the last task).
- The job folder uses approximately 25 GB.
- The largest files in the folder are the `args.json` files, which are 38557 (not so far from 4x9282=37128).
- Each one of these `args.json` file has a size of about 600K, and their total occupied size is then around 23 G. This the largest part of the whole job folder, then we'll ignore the rest.
- The different contribution to these 600K are listed below, using the approximate size estimates from https://pympler.readthedocs.io:

See script:
```python
import json
from pympler.asizeof import asizeof

FILENAME = "something.args.json"
with open(FILENAME, "r") as f:
    args = json.load(f)
args_size = asizeof(args)
print(f"args: {args_size}")
print("-" * 80)
for key, value in args.items():
    size_kB = asizeof(value) / 1000
    print(f"metadata    {key:22s}: size ~ {size_kB:07.2f} KB")
print("-" * 80)
for key, value in args["metadata"].items():
    try:
        len_value = len(value)
    except TypeError:
        len_value = 0
    size_kB = asizeof(value) / 1000
    print(f"metadata    {key:22s}: size ~ {size_kB:07.2f} KB (len={len_value})")
print("-" * 80)
```
and output:
```
args: 1092032
--------------------------------------------------------------------------------
metadata    input_paths           : size ~ 0000.21 KB
metadata    output_path           : size ~ 0000.12 KB
metadata    metadata              : size ~ 1089.17 KB
metadata    contrast_threshold    : size ~ 0000.03 KB
metadata    level                 : size ~ 0000.02 KB
metadata    overwrite             : size ~ 0000.03 KB
metadata    label_image_name      : size ~ 0000.06 KB
metadata    channel               : size ~ 0000.35 KB
metadata    label_image_cycle     : size ~ 0000.02 KB
metadata    intensity_image_cycle : size ~ 0000.03 KB
metadata    min_threshold         : size ~ 0000.03 KB
metadata    max_threshold         : size ~ 0000.03 KB
metadata    gaussian_blur         : size ~ 0000.03 KB
metadata    fill_holes_area       : size ~ 0000.03 KB
metadata    output_label_cycle    : size ~ 0000.03 KB
metadata    output_label_name     : size ~ 0000.06 KB
metadata    component             : size ~ 0000.10 KB
--------------------------------------------------------------------------------
metadata    plate                 : size ~ 0000.18 KB (len=1)
metadata    well                  : size ~ 0037.53 KB (len=357)
metadata    image                 : size ~ 1041.00 KB (len=9282)
metadata    num_levels            : size ~ 0000.03 KB (len=0)
metadata    coarsening_xy         : size ~ 0000.03 KB (len=0)
metadata    original_paths        : size ~ 0009.46 KB (len=26)
metadata    image_extension       : size ~ 0000.06 KB (len=3)
metadata    image_glob_patterns   : size ~ 0000.02 KB (len=0)
--------------------------------------------------------------------------------
```

It's clear that the `args["metadata"][["image"]` value (a list of 9282 strings) is the heaviest contributor to the file size.

After removing the image list, the size of one specific args.json file goes from 613K to 22K. This leads to a drastic size decrease of the args.json contribution, from about 23G to <1G.

tcompa:
Close #1237

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Unreleased
 
+* Runner:
+    * Do not include `metadata["image"]` in JSON file with task arguments (\#1238).
 * API:
     * Fix error message for rate limits in apply-workflow endpoint (\#1231).
 * Benchmarks:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,14 @@
 **Note**: Numbers like (\#123) point to closed Pull Requests on the fractal-server repository.
 
-# Unreleased
+# 1.4.7 (unreleased)
 
 * Runner:
     * Do not include `metadata["image"]` in JSON file with task arguments (\#1238).
+    * Add `FRACTAL_RUNNER_TASKS_INCLUDE_IMAGE` configuration variable, to define exceptions where tasks still require `metadata["image"]` (\#1238).
 * API:
     * Fix error message for rate limits in apply-workflow endpoint (\#1231).
 * Benchmarks:
-    * Add more scenarios, issue \#1184 (\#1232).
+    * Add more scenarios, as per issue \#1184 (\#1232).
 
 # 1.4.6
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 **Note**: Numbers like (\#123) point to closed Pull Requests on the fractal-server repository.
 
-# 1.4.7 (unreleased)
+# 1.4.7
 
 * Runner:
     * Do not include `metadata["image"]` in JSON file with task arguments (\#1238).
--- a/fractal_server/app/runner/_common.py
+++ b/fractal_server/app/runner/_common.py
@@ -18,7 +18,9 @@ from typing import Any
 from typing import Callable
 from typing import Optional
 
+from ...config import get_settings
 from ...logger import get_logger
+from ...syringe import Inject
 from ..models import Task
 from ..models import WorkflowTask
 from ..schemas import WorkflowTaskStatusType
@@ -26,6 +28,7 @@ from .common import JobExecutionError
 from .common import TaskExecutionError
 from .common import TaskParameters
 from .common import write_args_file
+
 
 HISTORY_FILENAME = "history.json"
 METADATA_FILENAME = "metadata.json"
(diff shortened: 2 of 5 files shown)


tcompa:
I ran fractal-containers with the last commit (3438a90), and both examples (01 and 02 from fractal-demos) do work.

Note that both examples include the copy-ome-zarr task:

image

tcompa:
This feature is currently in-place in this PR, but we need to know which and how many tasks will need an exception (that is, they will still require the `metadata["image"]` list).
In fractal-tasks-core, I could only find copy-ome-zarr.
@jluethi can you think of any other task (e.g. the 2D/3D auxiliary tasks package)?

As part of this PR, I hard-coded the fact that copy-ome-zarr represents an exception. This means that if we come up with other required exceptions we would need to stop the server and update to a new version.
If we expect multiple exceptions to appear in the near future, we can improve this procedure by exposing a configuration variable that defines the list of tasks that require this exception. In this way, the update of the list would only require a restart of fractal-server, and not a new release.

jluethi:
> @jluethi can you think of any other task (e.g. the 2D/3D auxiliary tasks package)?

Yes, the 2D to 3D task actually needs the same, good spot! See https://github.com/jluethi/napari-ome-zarr-roi-loader/blob/main/examples/convert_metadata_components_2D_to_3D.py
The task is called "Convert Metadata Components from 2D to 3D"

If we can easily do this via a config variable, that sounds good. But I'm fairly confident that only things switching between 2D & 3D currently use this metadata.

tcompa:
I've changed the task-exception-detection so that it now only uses task names (while in the first version I was also checking sources), and added a configuration variable
```python
FRACTAL_RUNNER_TASKS_INCLUDE_IMAGE: str = "Copy OME-Zarr structure;Convert Metadata Components from 2D to 3D"
```
which allows the admin to add more tasks if needed. Note: this procedure always requires a shutdown&restart of the fractal-server instance.

jluethi:
Nice! That approach looks good to me! :)

mfranzon (approved):

PR #1239, opened 2024-02-21
Fix wrong globbing filters within SLURM runner

### LINKED ISSUE #1240: Fix bug in slurm runner (glob is too broad)
If we have the list of components `["xxx_I_01_1", "xxx_I_01_11"]`, the `_glob_as_user` function incorrectly selects both files starting with `xxx_I_01_1` and with `xxx_I_01_11`. This becomes a bug when the function is expected to only select the `xxx_I_01_11` ones (for instance) and the list of `xxx_I_01_1` files is still changing (e.g. because the corresponding SLURM job is not over yet).

The fix in #1239 uses a more strict globbing function, which looks like
```python
def _glob_as_user_strict(
    *,
    folder: str,
    user: str,
    startswith: str,
) -> list[str]:
    """
    Run `ls` in a folder (as a user) and filter results
    Execute `ls` on a folder (impersonating a user, if `user` is not `None`)
    and select results that comply with a set of rules. They all start with
    `startswith` (if not `None`), and they match one of the known filename
    patterns.
    Arguments:
        folder: Absolute path to the folder
        user: If not `None`, the user to be impersonated via `sudo -u`
        startswith: If not `None`, this is used to filter output of `ls`.
    """

    res = _run_command_as_user(cmd=f"ls {folder}", user=user, check=True)
    output = res.stdout.split()

    new_output = []
    known_filenames = [
        f"{startswith}{suffix}"
        for suffix in [".args.json", ".metadiff.json", ".err", ".out"]
    ]
    for filename in output:
        if filename in known_filenames:
            new_output.append(filename)
        elif filename.startswith(f"{startswith}_out_") and filename.endswith(
            ".pickle"
        ):
            new_output.append(filename)

    return new_output
```

tcompa:
close #1240 

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,14 @@
 
 # 1.4.7
 
+This release provides a bugfix (PR 1239) and a workaround (PR 1238) for the
+SLURM runner, which became relevant for the use case of processing a large
+dataset (300 wells with 25 cycles each).
+
 * Runner:
     * Do not include `metadata["image"]` in JSON file with task arguments (\#1238).
     * Add `FRACTAL_RUNNER_TASKS_INCLUDE_IMAGE` configuration variable, to define exceptions where tasks still require `metadata["image"]` (\#1238).
+    * Fix bug in globbing patterns, when copying files from user-side to server-side job folder in SLURM executor (\#1239).
 * API:
     * Fix error message for rate limits in apply-workflow endpoint (\#1231).
 * Benchmarks:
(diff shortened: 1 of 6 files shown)


tcompa:
TODO: add unit test of the new aux function in `tests/05_backend/test_unit_custom_subprocess_module.py` (cc @mfranzon)

tcompa:
Just a confirmation, I ran this branch in fractal-containers and it shows no unexpected behavior:
```
$ cat fractal-containers/config.env 
# fractal-server
FRACTAL_SERVER_RELEASE=
FRACTAL_SERVER_GIT=debug-failed-cat-for-tmp-files

# fractal-client
FRACTAL_CLIENT_RELEASE=
FRACTAL_CLIENT_GIT=main

# fractal-tasks-core
FRACTAL_TASKS_CORE_RELEASE=
FRACTAL_TASKS_CORE_GIT=

# fractal-web
FRACTAL_WEB_GIT=

# fractal-demos
FRACTAL_DEMOS_GIT=
```


image

mfranzon (approved):

PR #1281, opened 2024-03-20
Benchmark images functions

tcompa:
## Checklist before merging


tcompa:
1. Unsurprisingly, the deduplicate function is slow.
2. TODO: We can skip using the deduplicate function in workflows with non-compound tasks, by modifying
     * merge_outputs function
     * updated_image block in runner.py
3. TBD: We probably will always need a deduplicate functionality within compound tasks, to sanitize the user-provided parallelization list. This has a worst-case (that is, when all parallelization items are different) quadratic scaling N*(N-1)/2. It could be made much faster by replacing object comparisons with string comparisons (together with JSON encoding/decoding), but we are not currently implementing this to avoid risks on unexpected behaviors.

PR #1282, opened 2024-03-21
Dev v2 post march 21

tcompa:
## Checklist before merging


tcompa:
Raw notes

Flags vs Types

-> no strong opinions
=> Let’s call them types
Improve naming

Current:
Dataset.flag_filters
WorkflowTask.flag_filters
Dataset.attribute_filters
WorkflowTask.attribute_filters

New options:
Dataset.filters.type
Dataset.filter_types

-> no strong opinion

=> we go with:
Dataset.filter_types
Dataset.filter_attributes
WorkflowTask.filters.types
WorkflowTask.filters.attributes
Changed images

Loop over changed-images
If path already exists -> EDIT IMAGE:
If origin is set:
If origin=path, this is irrelevant
If origin!=path, error?? => warning & ignore?
Higher level: what is the status of `origin`?
A property that tells the runner where to take “base” image metadata, and nothing more.
A required item in the task API, thus going back to input_path/output_path logic (-> not recommended)
A persistent information stored in the DB, then we should better define its specs.
What is the definition of origin?
How/when is it different from the `path` input argument?
Is it always present (apart from the “raw” images)?
Is it required in all task outputs? Are there defaults for when it’s missing?
It’s data provenance information that should also go into the OME-Zarr itself eventually. We may want to keep track of this server-side (for the metadata propagation)
Build changed image, by using the following sources (low- to high-priority):
Existing image (flags and attributes)
Task output (flags and attributes of the specific changed image)
NOTE: this must be optional, so that we also include the possibility of marking an image as “please update it”
New Dataset properties (flags and attributes, based on manifest and task output)
If path does not exist in the image list -> ADD NEW IMAGE:
Build new image, by using the following sources (low- to high-priority):
If origin exists and matches to an existing image, then use it as a lowest-priority image (flags and attributes).
Task output (flags and attributes of the specific changed image)
New Dataset filters (flags and attributes, based on manifest and task output)

Name:
changed_images
new_images
updated_images
image_list_updates

Removed:
image_list_removals: list[path]


New Task Output model:
image_list_updates: 
[{path=”/some/path”, origin=”something”, attributes={well=”B03”},}]
image_list_removals: [“/my/old/path”, “/my/old/path2”]
attributes: plate=”my_plate.zarr”
types: types={3D=True}

Attributes & types: Both the new filters & attributes applied to the image list entry
What goes in image_list_updates: Varies between images
What goes in attributes & types on their own: Global. Takes priority. Needs to be equal between all parallel images. Also goes into filters


New Task Output model V2:
image_list_updates: 
[{path=”/some/path”, origin=”something”, attributes={well=”B03”, plate=”my_plate.zarr”},types={3D=True}}]
image_list_removals: [“/my/old/path”, “/my/old/path2”]
Filters: 
Attributes: plate=”my_plate.zarr”
Types: types={3D=True}

Upside: Clearer separation of concerns
Downside: Some duplication of entries

=> let’s go with that!


Examples for some tasks:
Converter: 
Generates image_list_updates: some types (3D) & attributes (plate, well).
Filters: None?
Illum_corr:
Generates image_list_updates: type illumination_corrected = True
Filters: illumination_corrected = True
MIP:
Generates image_list_updates: new images with attribute wells same as before, but plate changing. type 3D=False
Filter: type 3D=False
Special version of MIP: Adds update to image list for original images to set their type 3D=True
Registration:
Apply registration works like illum_corr.
Cellpose & measurements:
No updates to image list, no new filters

tcompa:
Within the current PR, we already applied several updates based on a review of the runner internals.
Many of them will require similar updates for the API/web part - cc @ychiucco.

1. "flag" is replaced with "type", everywhere. For instance:
    * `Task.input_flags` goes to `Task.input_types` (and the same for `Task.output_flags`).
    * `SingleImage.flags` goes to `SingleImage.types`.
    * The `match_filters` arguments are renamed accordingly
2. `Dataset.attributes_filters` and `Dataset.flag_filters` are now `Dataset.filters.attributes` and `Dataset.filters.types` (@jluethi, I stand corrected with respect to yesterday: even if this option introduces one more nesting level it is more consistent with other uses of filters).
3. The same applies to `WorkflowTask.attribute_filters` and `WorkflowTask.flag_filters`, which become part of `WorkflowTask.filters`.
4. We slightly changed the output of `_find_single_image`, which now returns both the image and its positional index in the list (useful for when the image should be edited/removed). This is likely relevant within the images API.
5. We may also adapt `_filter_image_list` and `match_filter` further, so that they take a single `Optional[Filter]` object rather than separate attributes/types.

jluethi:
Sounds good! Fine by me to have the extra nesting layer of filters, it does look cleaner as well :)

PR #1416, opened 2024-04-18
Update `status` endpoint

### LINKED ISSUE #1083: WorkflowTask Status: When rerunning a workflow only partially, status of later tasks gets update late
If I have run a full workflow, then just run the beginning (e.g. until Maximum Intensity Projection), then the status still shows done for Cellpose & Napari workflows wrapper while the workflow is running again:
Screenshot 2023-12-13 at 18 18 07

And when it finishes, that status goes away.
Screenshot 2023-12-13 at 18 18 19

I would prefer it that it goes away immediately.

-----

On the other hand, when I then continue a workflow (e.g. run cellpose & napari workflows separately now), it correctly keeps showing the earlier statuses as done => as intended, that should stay this way!

mfranzon:
## Checklist before merging

closes #1083


fractal_server/app/routes/api/v2/dataset.py
```diff
@@ -303,7 +303,7 @@ async def get_workflowtask_status(
 
         # Highest priority: Read status updates coming from the running-job
         # temporary file. Note: this file only contains information on
-        # # WorkflowTask's that ran through successfully.
+        # WorkflowTask's that ran through successfully.
```
tcompa, line 306: This endpoint is now obsolete, right?

fractal_server/app/routes/api/v2/status.py
```diff
@@ -0,0 +1,147 @@
...
+
+from fastapi import APIRouter
+from fastapi import Depends
+from fastapi import HTTPException
+from fastapi import status
+
+from ....db import AsyncSession
+from ....db import get_async_db
+from ....models.v2 import JobV2
+from ....schemas.v2.dataset import DatasetStatusReadV2
```
tcompa, line 13: Can we move and rename this schema?

PR #1423, opened 2024-04-22
Workflow execution scaling

### LINKED ISSUE #1420: Workflow-execution scaling test
Take the current `test_fractal_demos_01`, and set the number of images (it's an argument of the first task) to N.

Modify the test so that `execute_tasks_v2` is run only once, for the entire workflow.

Let the test run, and report:

1. How many parallel or compound tasks it includes
2. How long it takes for the whole workflow to run
3. Check how many files exist in the job folder, and what is its size
4. (anecdotally): keep an eye on CPU usage (this will likely be large, since we are using ThreadPoolExecutor)
5. (optional, and only if using a script) use a profiler to identify the internal functions that contribute most to the total time

For N let's start with 10 and 100, and then see whether 1000 makes sense

(note that actual benchmarking should not take place in a test, but let's start from here and the iterate.. if this makes sense, we'll move to a script in an independent GitHub action)

ychiucco:
closes #1420 

## Checklist before merging

### DIFF
--- a/tests/v2/04_runner/test_fractal_examples.py
+++ b/tests/v2/04_runner/test_fractal_examples.py
@@ -931,3 +931,68 @@ def test_legacy_task(
         lines = f.read()
     assert 'SUBMIT 1-th task (name="dummy parallel")' in lines
     assert 'END    1-th task (name="dummy parallel")' in lines
+
+
+
+def test_scaling_workflow_execution(
+    tmp_path: Path, executor: Executor, fractal_tasks_mock_venv
+):
+    """
+    New mock of fractal-demos/examples/01.
+    """
+    N=10
+    WORKING_DIR = tmp_path / "job_dir"
+    zarr_dir = (tmp_path / "job_dir").as_posix().rstrip("/")
+    execute_tasks_v2_args = dict(
+        executor=executor,
+        workflow_dir=WORKING_DIR,
+        workflow_dir_user=WORKING_DIR,
+    )
+
+    
+    dataset_attrs = execute_tasks_v2(
+        wf_task_list=[
+            WorkflowTaskV2Mock(
+                task=fractal_tasks_mock_venv["create_ome_zarr_compound"],
+                args_non_parallel=dict(
+                    image_dir="/tmp/input_images", num_images=N
+                ),
+                args_parallel={},
+                id=0,
+                order=0,
+            ),
+            WorkflowTaskV2Mock(
+                task=fractal_tasks_mock_venv["illumination_correction"],
+                args_parallel=dict(overwrite_input=True),
+                id=1,
+                order=1,
+            ),
+            WorkflowTaskV2Mock(
+                task=fractal_tasks_mock_venv["MIP_compound"],
+                args_non_parallel=dict(suffix="mip"),
+                args_parallel={},
+                id=2,
+                order=2,
+            ),
+            WorkflowTaskV2Mock(
+                task=fractal_tasks_mock_venv["cellpose_segmentation"],
+                args_parallel={},
+                id=3,
+                order=3,
+            ),
+        ],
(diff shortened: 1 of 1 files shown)


tcompa:
- Move from test to standalone script (`executor` and `fractal_tasks_mock_venv` fixtures should be replaced with actual code)
- Report results in the form
    - Number N of images
    - Type/number of tasks
    - Total execution time
    - Total number of files in job folder
    - Total size of job folder
- Repeat for N=10,100 (as a start)
- Advanced: use a profiler to identify bottleneck functions/sections, if any
- Out of scope: set up a GHA that creates a diff for each PR

(all of this can live in `benchmarks/runner`, including benchmark results)

benchmarks/runner/benchmark_runner.py
```diff
@@ -0,0 +1,151 @@
...
+    return task_dict
+
+
+def benchmark(N: int):
+
+    tmp_path = tempfile.mkdtemp()
+    WORKING_DIR = Path(f"{tmp_path}/job_dir")
+    ZARR_DIR = (WORKING_DIR / "zarr").as_posix().rstrip("/")
+
+    fractal_tasks_mock_venv = mock_venv(tmp_path)
```
tcompa, line 89: Since this function is already well isolated, shall we move its call to the `main` block (and provide `fractal_tasks_mock_venv` as a function argument)?

benchmarks/runner/benchmark_runner.py
```diff
@@ -0,0 +1,193 @@
...
+
+    for N in range(100, 1001, 100):
+
+        tmp_path = tempfile.mkdtemp()
+        cProfile.run(f"benchmark({N}, '{tmp_path}')", "profile_results")
+        stats = pstats.Stats("profile_results")
+        stats.sort_stats("tottime")
+
+        thread_time = stats.stats[
+            "~", 0, "<method 'acquire' of '_thread.lock' objects>"
```
tcompa, line 134: Let's review together that this is the right approach (with an artificial example).

benchmarks/runner/benchmark_runner.py
```diff
@@ -0,0 +1,193 @@
...
+        ][2]
+        total_time = stats.total_tt
+
+        list_dirs = {}
+        for path, key in [
+            (tmp_path, "base"),
+            (f"{tmp_path}/job_dir", "job_dir"),
```
tcompa, line 141: only job-dir is needed

benchmarks/runner/benchmark_runner.py
```diff
@@ -0,0 +1,193 @@
...
+
+    runner = os.path.dirname(benchmarks.runner.__path__[0])
+    with open(f"{runner}/runner/runner_benchmark.txt", "w") as file:
+        # Headers
+        to_write = (
+            "\n\n\n"
+            "|\tN\t"
+            "|\tthread\t"
+            "|\ttotal\t"
+            "|\tthread/total\t"
```
tcompa, line 171: difference would be more informative than ratio here

benchmarks/runner/benchmark_runner.py
```diff
@@ -0,0 +1,193 @@
...
+            dict(
+                N=N,
+                thread_time=thread_time,
+                total_time=total_time,
+                list_dirs=list_dirs,
+            )
+        )
+
+    runner = os.path.dirname(benchmarks.runner.__path__[0])
+    with open(f"{runner}/runner/runner_benchmark.txt", "w") as file:
```
tcompa, line 164: Let's include a small preamble, before the table, saying

> Type/number of tasks

benchmarks/runner/benchmark_runner.py
ychiucco: https://github.com/fractal-analytics-platform/fractal-server/issues/1420#issuecomment-2074920560

PR #1454, opened 2024-05-01
Remove `fractal_tasks_mock` from `test_non_executable_task_command`

### LINKED ISSUE #1377: Re-enable `test_non_executable_task_command`
Relevant file: tests/v2/08_full_workflow/test_full_workflow_v2.py

As of #1371, the test is up-to-date, and runs through for the local backend. If fails with the slurm backend for some test-related issue (`RuntimeError: This event loop is already running`).

Side-comment: including the (unrelated) `fractal_tasks_mock` fixture makes the test pass, but it's not clear why.

ychiucco:
closes #1377 

## Checklist before merging


tests/v2/08_full_workflow/common/common.py
```diff
@@ -17,49 +14,17 @@ def _task_name_to_id(task_name: str, task_list: list[dict[str, Any]]) -> int:
...
     project_factory_v2,
-    dataset_factory_v2,
     workflow_factory_v2,
-    backend,
-    override_settings_factory,
-    tmp_path_factory,
-    fractal_tasks_mock,
-    request,
+    dataset_factory_v2,
+    user_kwargs: dict = {},
```
tcompa, line 23: Let's avoid mutable default arguments.
The typical solution here is `Optional[dict] = None`, and then handle the `None` from the function code.

tests/v2/08_full_workflow/test_full_workflow_local_v2.py
```diff
@@ -0,0 +1,128 @@
+import pytest
+from common.common import failing_workflow_UnknownError
```
tcompa, line 2: Why a nested `common.common` structure? Could we simply use `common.py`?
(maybe with a more clear name like `common_functions`)

tests/v2/08_full_workflow/test_full_workflow_local_v2.py
```diff
@@ -0,0 +1,128 @@
...
+    dataset_factory_v2,
+    workflow_factory_v2,
+    override_settings_factory,
+    tmp_path_factory,
+    fractal_tasks_mock,  # needed
+):
+    # Use a session-scoped FRACTAL_TASKS_DIR folder
+    override_settings_factory(
+        FRACTAL_RUNNER_BACKEND="local",
+        FRACTAL_RUNNER_WORKING_BASE_DIR=tmp777_path / "artifacts-local",
```
tcompa, line 22: No need for `artifacts-local` any more, since this test is not parametrized over backend. We can keep it cleaner, with just `artifacts`.
(same applies to most other tests)

tests/v2/08_full_workflow/test_full_workflow_local_v2.py
```diff
@@ -0,0 +1,128 @@
...
+    project_factory_v2,
+    dataset_factory_v2,
+    workflow_factory_v2,
+    override_settings_factory,
+    tmp_path_factory,
+    fractal_tasks_mock,  # needed
+):
+    # Use a session-scoped FRACTAL_TASKS_DIR folder
+    override_settings_factory(
+        FRACTAL_RUNNER_BACKEND="local",
```
tcompa, line 21: Can we set this as a module-level variable? It's be useful to avoid typos in one of the test.

PR #1458, opened 2024-05-03
Return a JSON payload for 500-Internal-Server-Error

### LINKED ISSUE #1406: Return a JSON payload also for 500 errors
Currently the API return the text "Internal Server Error" when there is a 500 error. The frontend expects that all the API calls return a JSON response and attempts to parse the response as JSON, resulting in a misleading `SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data`.

This can be avoided if the server returns a JSON payload also in case of 500 errors.

The payload could be something like:

```json
{"detail":"Internal Server Error"}
```

ychiucco:
closes #1406 

## Checklist before merging

### DIFF
--- a/fractal_server/main.py
+++ b/fractal_server/main.py
@@ -120,7 +120,7 @@ async def on_startup() -> None:
 
 
 @app.middleware("http")
-async def return_JSON_if_500(request: Request, call_next):
+async def internal_server_error_middleware(request: Request, call_next):
     try:
         response = await call_next(request)
     except Exception as e:
--- a/fractal_server/main.py
+++ b/fractal_server/main.py
@@ -16,6 +16,8 @@
 This module sets up the FastAPI application that serves the Fractal Server.
 """
 from fastapi import FastAPI
+from fastapi import Request
+from fastapi.responses import JSONResponse
 
 from .app.security import _create_first_user
 from .config import get_settings
@@ -115,3 +117,18 @@ async def on_startup() -> None:
         is_verified=True,
     )
     await __on_startup()
+
+
+@app.middleware("http")
+async def return_JSON_if_500(request: Request, call_next):
+    try:
+        response = await call_next(request)
+    except Exception as e:
+        response = JSONResponse(
+            status_code=500,
+            content={
+                "detail": "Internal server error occurred",
+                "original_error": str(e),
+            },
+        )
+    return response
--- a/fractal_server/app/routes/api/__init__.py
+++ b/fractal_server/app/routes/api/__init__.py
@@ -17,8 +17,3 @@ async def alive():
         alive=True,
         version=settings.PROJECT_VERSION,
     )
-
-
-@router_api.get("/bug/")
-async def bug():
-    raise Exception("BUG!")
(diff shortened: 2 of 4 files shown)


fractal_server/main.py
```diff
@@ -115,3 +117,18 @@ async def on_startup() -> None:
         is_verified=True,
     )
     await __on_startup()
+
+
+@app.middleware("http")
```
tcompa, line 122: what does http mean here?

ychiucco: FastAPI doc: https://fastapi.tiangolo.com/tutorial/middleware/#create-a-middleware
FastAPI source code: https://github.com/tiangolo/fastapi/blob/1c3e6918750ccb3f20ea260e9a4238ce2c0e5f63/fastapi/applications.py#L4503

ychiucco:
From https://www.starlette.io/middleware/#using-middleware
```
Every Starlette application automatically includes two pieces of middleware by default:
* ServerErrorMiddleware - Ensures that application exceptions may return a custom 500 page, or display an application traceback in DEBUG mode. This is always the outermost middleware layer.
* ExceptionMiddleware - Adds exception handlers, so that particular types of expected exception cases can be associated with handler functions. For example raising HTTPException(status_code=404) within an endpoint will end up rendering a custom 404 page.
```

We are actually already using ServerErrorMiddleware: 
```
/Users/yuri/Library/Caches/pypoetry/virtualenvs/fractal-server-t2lE8eOX-py3.10/lib/python3.10/site-packages/starlette/middleware/errors.py:186: in __call__
    raise exc
/Users/yuri/Library/Caches/pypoetry/virtualenvs/fractal-server-t2lE8eOX-py3.10/lib/python3.10/site-packages/starlette/middleware/errors.py:164: in __call__
    await self.app(scope, receive, _send)
```

But it seems that the intention is to continue to raise the error (https://github.com/encode/starlette/blob/9f16bf5c25e126200701f6e04330864f4a91a898/starlette/middleware/errors.py#L183)
```python
# We always continue to raise the exception.
# This allows servers to log the error, or allows test clients
# to optionally raise the error within the test case.
raise exc
```

ychiucco:
From https://www.starlette.io/middleware/#using-middleware
```
Every Starlette application automatically includes two pieces of middleware by default:
* ServerErrorMiddleware - Ensures that application exceptions may return a custom 500 page, or display an application traceback in DEBUG mode. This is always the outermost middleware layer.
* ExceptionMiddleware - Adds exception handlers, so that particular types of expected exception cases can be associated with handler functions. For example raising HTTPException(status_code=404) within an endpoint will end up rendering a custom 404 page.
```

We are actually already using ServerErrorMiddleware: 
```
/Users/yuri/Library/Caches/pypoetry/virtualenvs/fractal-server-t2lE8eOX-py3.10/lib/python3.10/site-packages/starlette/middleware/errors.py:186: in __call__
    raise exc
/Users/yuri/Library/Caches/pypoetry/virtualenvs/fractal-server-t2lE8eOX-py3.10/lib/python3.10/site-packages/starlette/middleware/errors.py:164: in __call__
    await self.app(scope, receive, _send)
```

But it seems that the intention is to continue to raise the error (https://github.com/encode/starlette/blob/9f16bf5c25e126200701f6e04330864f4a91a898/starlette/middleware/errors.py#L183)
```python
# We always continue to raise the exception.
# This allows servers to log the error, or allows test clients
# to optionally raise the error within the test case.
raise exc
```

PR #1480, opened 2024-05-13
Conditionally include V1 API

### LINKED ISSUE #1477: Conditionally include V1 API
Let's add a new configuration variable
```python
FRACTAL_API_V1_ENABLED: Literal["true", "false"] = "true"
"""
Whether to include the v1 API; accepted values are `true` (default) or `false`.
"""
```
such that if `FRACTAL_API_V1_ENABLED == "false"` then we do not include the V1 routers.

(even better if can stick with a `bool` variable, but we need to make sure of which string values are parsed and coerced correctly).

ychiucco:
closes #1477 

## Checklist before merging


ychiucco:
https://github.com/fractal-analytics-platform/fractal-server/issues/1477#issuecomment-2107096875

PR #1489, opened 2024-05-14
Clean up `scripts` folders

### LINKED ISSUE #1485: Clean up scripts folders
* Remove `scripts/database_updates`
* Split `scripts/populate_db` into `client` (which only includes an init file and `fractal_client.py`, with the latter renamed `client.py`)
* Move the populate_db Python and bash scripts into `benchmarks/populate_db` (including the env file).
* Add README to `benchmarks/runner`
* Move `scripts/fractal_service` and `scripts/mypy-scan` and `scripts/start_server_with_gunicorn.sh` to `scripts/old`, with a README clarifying that those are unsupported examples.

ychiucco:
closes #1485 

## Checklist before merging


examples/example_1.py
```diff
@@ -4,8 +4,8 @@
 
 sys.path.append((Path(__file__).parents[1] / "scripts/populate_db").as_posix())
```
tcompa, line 5: This line should be updated

PR #1500, opened 2024-05-15
Set short timeout for `docker compose down`

tcompa:
## Checklist before merging


tcompa:
For the record, our tests with @mfranzon suggest that the reason why we always hit the timeout is that the docker entrypoint (this one: https://github.com/rancavil/slurm-cluster/blob/devel/master/docker-entrypoint.sh) is not stopping correctly when receiving a SIGTERM.

Possibly related:
* https://vsupalov.com/docker-compose-stop-slow/
* https://hynek.me/articles/docker-signals/


We are not investigating this further for the moment, since this PR does not really change the behavior (we were anyway already waiting for the teardown timeout, even if it was 10 seconds rather than 1 second).

PR #1524, opened 2024-05-27
Remove `Project[V2].user_list`

### LINKED ISSUE #1503: Replace use of `Project.user_list` and `ProjectV2.user_list` with `join`s
This is not meant to fix #1463, but rather to set up a more controlled testing environment.

ychiucco:
closes #1503 

## Checklist before merging


fractal_server/app/routes/api/v1/project.py
```diff
@@ -85,12 +85,11 @@ async def create_project(
...
 
     db_project = Project(**project.dict())
-    db_project.user_list.append(user)
+
     try:
         db.add(db_project)
         await db.commit()
         await db.refresh(db_project)
-        await db.close()
     except IntegrityError as e:
```
tcompa, line 93: I don't think we use this pattern (catching `IntegrityError` and transforming it from 500 to 422) often - to be checked. We can probably drop it from this endpoint.

Let's make these 500 errors evident, and then if it's relevant we should turn this into a helper function and use it everywhere to wrap the `db.commit` statements from endpoints.

ychiucco: We catch `IntegrityError`s just in `app/routes/api/v1/project.py::create_project` and `app/security/__init__.py::_create_first_user`.

ychiucco: Closed with https://github.com/fractal-analytics-platform/fractal-server/pull/1530

PR #1545, opened 2024-06-04
New SSH features (task collection and job execution)

### LINKED ISSUE #1580: Fix wrong use of `rstrip` in ssh executor
Let's take a note here, since this one already bit us in the past.. When passing a multi-character string to `rstrip`, the behavior is not necessarily the expected one:

> Return a copy of the string with trailing characters removed. The chars argument is a string specifying the set of characters to be removed. If omitted or None, the chars argument defaults to removing whitespace. The chars argument is not a suffix; rather, all combinations of its values are stripped:
> (https://docs.python.org/3/library/stdtypes.html#str.rstrip)

which means
```python repl
>>> print("aaa-zarr.tar.gz".rstrip(".tar.gz"))
aaa-
```

(thanks @jluethi for reporting)

tcompa:
close #1580

## Checklist before merging


tcompa:
Here is a plan for the review by @mfranzon

CHANGES TO REVIEW
* config.py: new SSH variables (to be improved: https://github.com/fractal-analytics-platform/fractal-server/issues/1592)
    * credentials
    * backend_runner slurm_ssh
    * remote working directory
* app/routes:
    * small refactor of is-shutdown-available internal functions
    * new task-collection endpoints and a conditional inclusion of their router
* Split of app/runner/executors/slurm/sudo vs app/runner/executors/slurm/ssh
* Actual runner app/runner/v2/_slurm_ssh/ -> review background-task interface (e.g. new `connection`)
* Lifespan -> FastAPI app object has `.state.connection` attribute (the fabric SSH connection), prepared during lifespan setup phase, and propagated from endpoints to internal functions (including job execution)
* fractal_server/ssh -> all current fabric logic
* support module for compressing folders:
    * FRACTAL_SLURM_WORKER_PYTHON -m fractal_server.app.runner.compress_folder (this includes the two versions, tarfile vs tar) (this also includes the copy-tar-remove workaround for performance)
* task collection app/tasks/v2/*ssh* (bash template scripts)
    * there is a test for the happy path
    * failure handling in background task is in principle ok
    * to review: possible fail-fast situations in endpoint (422)
* preliminary tests in tests/v2/00_ssh or tests/v2/08_full_workflow/test_full_workflow_slurm_ssh_v2.py


NO REVIEW
* docs: relative paths changed
* tests/v2/fractal_tasks_mock: replaced print with logging
* SSH executor details: this has to be cleaned up and tested elsewhere

mfranzon (approved):

PR #1547, opened 2024-06-04
add shutdown logic in lifespan for executors

### LINKED ISSUE #1519: Trigger SLURM executor shutdown when worker process receives SIGABRT
Branching from https://github.com/fractal-analytics-platform/fractal-server/issues/1499, where we introduce the sigabrt-catching for each worker process (that is, for each app). As a first step, we're converting SIGABRT in SIGTERM - so that the app-shutdown logic is executed. This issue concerns the actual shutdown logic.

## Option 1 (not recommended)
We wait. If I remember correctly, the first SIGTERM would wait for all background jobs to finish.
In principle this would work, but I think that the worker would not be restarted until it's completely shut down. During this (possibly long) time interval, there would be one less worker available. If for whatever reason all current workers have a timeout error, then the API would not be responsive for as long as the SLURM job will take.

## Option 2 (recommended)
We trigger the SLURM-executor shutdown for all jobs associated to this worker process (requires #1504), wait a certain time interval, and exit. The code here would look like:
```python
# If backend is not SLURM, skip this part
# Note that this would need to take place for both V1 and V2 jobs.
jobs = (
    await db.execute(
        select(JobV2)
            .where(JobV2.id in app.state.jobs)
            .where(JobV2.status == "submitted")
            )
            ).scalars().all()  # untested  
for job in jobs:
    _write_shutdown_file(job=job)

t_start = time.perf_counter()
while time.perf_counter() - t_start < FRACTAL_GRACEFUL_SHUTDOWN_TIME:  # this could be e.g. 30 seconds
    print("Waiting 3 seconds before checking")
    time.sleep(3)
    jobs = (
        await db.execute(
            select(JobV2)
                .where(JobV2.id in app.state.jobs)
                .where(JobV2.status == "submitted")
                )
            ).scalars().all()  # untested  
    if len(jobs) == 0:
        print("All jobs associated to this app are either done or failed. Exit.")
        break
    else
        print("Some jobs are still 'submitted' (V1 IDs: x, y, z, V2 IDs: x, y, z))")
print("Graceful shutdown reached its maximum time, but some jobs are still submitted")
jobs = (
    await db.execute(
        select(JobV2)
            .where(JobV2.id in app.state.jobs)
            .where(JobV2.status == "submitted")
            )
        ).scalars().all()  # untested  
for job in jobs:
    job.status = "failed"
    if job.log is None:
        job.log = "Job stopped due to app shutdown\n"
    else:
        job.log += "Job stopped due to app shutdown\n"
    db.add(job)
    await db.commit()
print("Exit from shutdown logic")
```

mfranzon:
## Checklist before merging

close #1519

### DIFF
--- a/fractal_server/app/runner/shutdown.py
+++ b/fractal_server/app/runner/shutdown.py
@@ -13,10 +13,6 @@ from fractal_server.logger import get_logger
 from fractal_server.syringe import Inject
 
 
-# If backend is not SLURM, skip this part
-# Note that this would need to take place for both V1 and V2 jobs.
-
-
 async def cleanup_after_shutdown(
     jobsV1: list[int], jobsV2: list[int], logger_name: str
 ):
--- a/fractal_server/app/runner/shutdown.py
+++ b/fractal_server/app/runner/shutdown.py
@@ -1,5 +1,4 @@
 import time
-from logging import Logger
 
 from sqlmodel import select
 
@@ -10,6 +9,7 @@ from fractal_server.app.models.v2 import JobV2
 from fractal_server.app.models.v2.job import JobStatusTypeV2
 from fractal_server.app.routes.aux._job import _write_shutdown_file
 from fractal_server.config import get_settings
+from fractal_server.logger import set_logger
 from fractal_server.syringe import Inject
 
 
@@ -18,8 +18,10 @@ from fractal_server.syringe import Inject
 
 
 async def cleanup_after_shutdown(
-    jobsV1: list[int], jobsV2: list[int], logger: Logger
+    jobsV1: list[int], jobsV2: list[int], logger_name: str
 ):
+    logger = set_logger(logger_name)
+    logger.propagate = True
     logger.info("Cleanup function after shutdown")
     stm_v2 = (
         select(JobV2)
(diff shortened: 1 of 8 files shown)


fractal_server/app/runner/shutdown.py
```diff
@@ -0,0 +1,161 @@
...
+                            select(ApplyWorkflow)
+                            .where(ApplyWorkflow.id.in_(jobsV1))
+                            .where(ApplyWorkflow.status == "submitted")
+                        )
+                    )
+                    .scalars()
+                    .all()
+                )
+
+                for job in jobsV2_db:
```
tcompa, line 85: I think these additional calls to `_write_shutdown_file` are not needed. The file is written once, at the beginning, and then we are just waiting for when the executor identifies it and triggers the rest of its own shutdown (so that the background task also marks jobs as failed in the database).

fractal_server/app/runner/shutdown.py
```diff
@@ -0,0 +1,161 @@
...
+                        .where(ApplyWorkflow.id.in_(jobsV1))
+                        .where(ApplyWorkflow.status == "submitted")
+                    )
+                )
+                .scalars()
+                .all()
+            )
+
+            for job in jobsV2_db:
+                _write_shutdown_file(job=job)
```
tcompa, line 137: same as above, I think we should not write shutdown files any more

PR #1662, opened 2024-07-17
New `/settings/` endpoint

### LINKED ISSUE #1659: Expose `/api/settings/` endpoint
This should be restricted to superusers, and it should obfuscate/remove any sensitive configuration variable.

ychiucco:
closes #1659 

## Checklist before merging

### DIFF
--- a/fractal_server/app/routes/api/__init__.py
+++ b/fractal_server/app/routes/api/__init__.py
@@ -25,7 +25,4 @@ async def alive():
 @router_api.get("/settings/")
 async def view_settings(user: UserOAuth = Depends(current_active_superuser)):
     settings = Inject(get_settings)
-    for k in settings.dict():
-        if ("PASSWORD" in k) or ("SECRET" in k):
-            setattr(settings, k, "***")
-    return settings
+    return settings.get_sanitized()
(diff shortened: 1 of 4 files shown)


tcompa:
How does this look when we have an OAuth client: are we ever printing their SECRET?
The reason I ask is that the key (`OAUTH_CLIENTS_CONFIG`) does not include the SECRET or PASSWORD strings.

fractal_server/app/routes/api/__init__.py
```diff
@@ -17,3 +20,12 @@ async def alive():
         alive=True,
         version=settings.PROJECT_VERSION,
     )
+
+
+@router_api.get("/settings/")
+async def view_settings(user: UserOAuth = Depends(current_active_superuser)):
+    settings = Inject(get_settings)
+    for k in settings.dict():
+        if ("PASSWORD" in k) or ("SECRET" in k):
```
tcompa, line 29: Let's use `k.upper()`

tcompa: And let's include TOKEN and PWD

tcompa (approved):

PR #1675, opened 2024-07-23
Restrict to `pip<24.1`

### LINKED ISSUE #1591: Pin task-collection `pip` to `<24.1`?
As of `pip` v24.1, the scmultiplex task package cannot be installed (ref https://github.com/fmi-basel/gliberal-scMultipleX/issues/106).

Note that task collection always runs `pip install --upgrade pip` before installing a package, which means that any task-collection of that package taking place now will likely fail, even without a version update of fractal-server.

The specific issue is related to using the 1.4 prefect version, which has some old-style metadata which are now deprecated as of pip=24.1 (see https://pip.pypa.io/en/stable/news/#deprecations-and-removals and https://github.com/pypa/pip/issues/12063). I don't know whether other task packages are affected.

I have no idea of how complex it would be to fix the specific scmultiplex issue, but for the moment the only fractal-side fix is to constrain the task-collection pip to `<24.1`.

ychiucco:
closes #1591 

## Checklist before merging


tcompa:
Let's modify one of the task collection tests, setting the max version e.g. to 23 and adding an assertion on the pip version.
Note that an easy way to do this is to access `collection_state.data["freeze"]` and look for the `pip=...` entry.

PR #1738, opened 2024-09-09
CRUD of user groups

### LINKED ISSUE #1732: Create default user group
Requires
* #1731

* A default user group should be always present in the db, and all users should be associated to it.
* The name can be e.g. "all" or "all-users".
* If possible, the check and creation should be part of the `set-db` command, rather than taking place at every application startup.

### LINKED ISSUE #1733: Move first-user creation to `set-db`
Similar to #1732.

Right now, at every application startup we run the following logic:
1. Check whether at least one superuser exists
2. If not, create a default one in the database

This is suboptimal, for at least two reasons:
* When e.g. N=12 workers are started for the first time, they all perform the same operations - and therefore we need to catch integrity errors when they all try to create a new user
* In a live instance, default-user creation should happen only once, but we are still performing N checks at each startup even after a long uptime

As in #1732: if possible, let's move steps 1 and 2 above into `set-db` logic

tcompa:
Close #1732
Close #1733

## Checklist before merging


fractal_server/app/routes/auth/group.py
```diff
@@ -0,0 +1,173 @@
...
+@router_group.get(
+    "/group/", response_model=list[UserGroupRead], status_code=200
+)
+async def get_list_user_groups(
+    user_ids: bool = False,
+    user: UserOAuth = Depends(current_active_superuser),
+    db: AsyncSession = Depends(get_async_db),
+) -> list[UserGroupRead]:
+    """
+    FIXME docstring
```
ychiucco, line 35: add docstring

fractal_server/app/routes/auth/group.py
```diff
@@ -0,0 +1,173 @@
...
+    status_code=status.HTTP_200_OK,
+)
+async def update_single_group(
+    group_id: int,
+    group_update: UserGroupUpdate,
+    user: UserOAuth = Depends(current_active_superuser),
+    db: AsyncSession = Depends(get_async_db),
+) -> UserGroupRead:
+    """
+    FIXME docstring
```
ychiucco, line 123: add docstring

fractal_server/app/routes/auth/group.py
```diff
@@ -0,0 +1,173 @@
...
+@router_group.delete(
+    "/group/{group_id}/", status_code=status.HTTP_405_METHOD_NOT_ALLOWED
+)
+async def delete_single_group(
+    group_id: int,
+    user: UserOAuth = Depends(current_active_superuser),
+    db: AsyncSession = Depends(get_async_db),
+) -> UserGroupRead:
+    """
+    FIXME docstring
```
ychiucco, line 165: add docstring

fractal_server/app/routes/auth/group_names.py
```diff
@@ -0,0 +1,33 @@
...
+
+@router_group_names.get(
+    "/group-names/", response_model=list[str], status_code=status.HTTP_200_OK
+)
+async def get_list_user_group_names(
+    user: UserOAuth = Depends(current_active_user),
+    db: AsyncSession = Depends(get_async_db),
+) -> list[str]:
+    """
+    FIXME docstring
```
ychiucco, line 26: add docstring

ychiucco (approved):

PR #1765, opened 2024-09-12
Test OAuth flow after email update

### LINKED ISSUE #1758: Review OAuth flow after email update
Branching from #1756

We should introduce another script like https://github.com/fractal-analytics-platform/fractal-server/blob/main/scripts/oauth/oauth.sh, where the expected behavior is like this (to be confirmed):

1. Repeat all what is in the current scritp, so that a Fractal user exists and is associated to an oauth account.
2. Change the user email (easiest done via psql, otherwise via fractal-client - TBD)
3. Repeat all relevant steps to reach the GET-current-user, to verify that the oauth account can still be used to access the Fractal user (note: there must be a single fractal user).
4. Manually remove the `oauth_account` related to the given `user_oauth`
5. Repeat all steps as in item 3, and verify that oauth access to fractal is still working but it creates a new fractal user.

If all works as described above, then we don't need to prevent email changes - even in the presence of oauth accounts.
The reason is that only a set of deliberate operations would remove access to a given fractal account, that is, changing their email **and** removing their oauth account from the database (which is not even exposed in the API).

Ref: https://fastapi-users.github.io/fastapi-users/latest/configuration/oauth/?h=assoc#existing-account-association

ychiucco:
closes #1758 

## Checklist before merging
- [x] Add unit test


ychiucco:
https://github.com/fractal-analytics-platform/fractal-server/pull/1766#issuecomment-2346659431

fractal_server/app/schemas/user.py
```diff
@@ -20,6 +20,13 @@
 )
 
 
+class OAuthAccountRead(BaseModel):
```
tcompa, line 23: Docstring

fractal_server/app/schemas/user.py
```diff
@@ -37,6 +44,7 @@ class UserRead(schemas.BaseUser[int]):
     slurm_accounts: list[str]
     group_names: Optional[list[str]] = None
     group_ids: Optional[list[int]] = None
+    oauth_accounts: list[OAuthAccountRead] = Field(default_factory=list)
```
tcompa, line 47: Is the default needed?

fractal_server/app/routes/auth/users.py
```diff
@@ -189,9 +188,11 @@ async def list_users(
     # https://github.com/fractal-analytics-platform/fractal-server/issues/1742
     for ind, user in enumerate(user_list):
         user_list[ind] = dict(
-            user.model_dump(),
+            **user.model_dump(),
+            oauth_accounts=[account for account in user.oauth_accounts],
```
tcompa, line 192: Can we use `user.oauth_accounts` directly?

PR #1811, opened 2024-09-25
[DO NOT MERGE] Deprecate full-support for V1

tcompa:
This is not (yet) to merge, but mostly to verify how much coverage we'd be missing after removing the relevant code/tests.

## Checklist before merging


tcompa:
As partly expected, the naive approach of removing all obsolete V1 tests is not acceptable, as some of those tests are the only ones covering some parts of the codebase which is shared by V1 and V2.

Closing this PR, we'll need to start afresh and be more careful in migrating V1 tests into V2 form.

Issue #1803, opened 2024-09-24
Major update of tasks


tcompa:
> Feedback will be welcome, but we are currently still working on this plan.

---

The plan below aims at reaching these high-level goals:

1. It introduces `TaskGroupV2` collections of `TaskV2`s.
2. It fully revamps the access-control rules for tasks and task groups, going from the current `owner/username` matching to using database relationships.
3. It makes it possible to non-disruptively revoke a user's access to a task, even after the user already used that task (ref #1744).

A few parts are still partly undefined, either because they are TBD or because we'd rather postpone them to a different update. The main undefined areas are:

4. Task lifecycle - this is clearly postponed to a different update.
5. How are attributes precisely distributed into `TaskV2` vs `TaskGroupV2`
    - [ ] #1828
    - [ ] #1829
    - [ ] #1831
6. How is the identity of tasks or task groups defined, in two areas:
    a. What kind of checks prevent/allow two independent collections of the same task package by two different users? Are they based on folder paths, on `TaskGroupV2` metadata, or on something else?
    b. Is the `TaskV2.source` still a suitable attempt towards a unique task identifier? This is relevant both for point `a` above and for how to make tasks portable across different fractal-server instances.
7. Will we able to fully drop/upgrade the current `owner/username` access-control logic? Or do we need to remain backwards compatible.
8. How will we handle v1 tasks? Ideally we'd like to not modify teir data structure and access-control rules at all; if this would introduce too much friction, then we'll have to make changes to V1 tasks as well.

# Database structure

## Tasks and task groups

- [x] We add a new model `TaskGroupV2`, with its new table `taskgroupv2`.
- [x] We include an _optional_ foreign key `taskv2.taskgroupv2_id` -> this means that all tasks must belong to a group (even if the group size is one).
- [x] ~~FUTURE RELEASE: We make the FK required~~ POSTPONED - ref #1837
- [x] We include a ORM attribute `TaskGroupV2.task_list`, with cascade delete (deleting a task group also deletes all its tasks).

## Tasks and users or user groups

- [x] Required foreign key `taskgroupv2.user_id`.
- [x] Optional foreign key `taskgroupv2.user_group_id`.
- [x] ~~Test that when a `UserGroup` is deleted, all corresponding `taskgroupv2.user_group_id` values are set to `None`.~~ Show that you cannot delete a `UserGroup` if it is linked to a `TaskGroupV2`, but you have to manually perform "cascade" operations and set FKs to None.

[...]

PR #1817, opened 2024-09-27
Introduce `TaskGroup`s

ychiucco:
ref #1803

## Checklist before merging
- [ ] I reviewed all FIXMEs


tcompa:
The core changes to database models are now in.

The next step is to make sure that all TaskV2 objects in the CI conform to the new models:

1. The `task_factory_v2` should create a dummy `TaskGroupV2` for each task that it creates. Note that we need to add an argument `user_id` to `task_factory_v2`, to be used when creating the task group.
2. The single-task POST endpoint should create a `TaskGroupV2` entry and establish the association with the new `TaskV2`
   * user_id is the one of the user making the API request
   * user_group_id is the one of the All group - for the moment
   * user_group_id is an request-body parameter - TBD
3. The same applies to the two (local/ssh) task-collection endpoints
4. The same applies to the custom task-collection endpoint

fractal_server/app/routes/auth/_aux_auth.py
```diff
@@ -111,3 +113,46 @@ async def _user_or_404(user_id: int, db: AsyncSession) -> UserOAuth:
...
+    res = await db.execute(stm)
+    user_group_id = res.scalars().one_or_none()
+    if user_group_id is None:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"User group '{FRACTAL_DEFAULT_GROUP_NAME}' not found.",
+        )
+
+
+def _get_default_user_group_id_sync(db: DBSyncSession) -> UserGroup:
```
tcompa, line 131: check if this is used

fractal_server/app/models/v2/task.py
```diff
@@ -39,3 +44,25 @@ class TaskV2(SQLModel, table=True):
...
+
+
+class TaskGroupV2(SQLModel, table=True):
+
+    id: Optional[int] = Field(default=None, primary_key=True)
+
+    user_id: int = Field(foreign_key="user_oauth.id")
+    user_group_id: Optional[int] = Field(foreign_key="usergroup.id")
+
+    active: bool
```
tcompa, line 58: add default True

PR #1856, opened 2024-10-07
New `TaskV2` and `TaskGroupV2` attributes

### LINKED ISSUE #1828: Introduce new TaskV2 attributes
We should enrich the task model with some new (optional) attributes:
1. `type: Optional[str] = None`. Examples of typical values: `Conversion`, `Segmentation`, `Registration`, `Measurement`
    * `type` is already in use for parallel/non-parallel/compound tasks - let's find a better name
2. `modality: Optional[str] = None`. Examples of typical values: `HCS`, `lightsheet`, `EM`
3. `authors: list[str] = Field(default_factory=list)`. Example of typical values: `["Name1 Surname1", "Name2 Surname2"]` or `["Organization 1"]`
4. `tags: list[str] = Field(default_factory=list)`.

### Changes to make

Planned changes include

- [x] Update the database `TaskV2` model and create a new schema-migration script
- [x] Update the single-task API schemas (Create/Update/Read)
- [x] Update the task-collection manifest schema
- [x] Propagate single-task attributes from the manifest into the single task, upon task creation during task collection
- [x] Same as above, for the "global" `authors` attribute
- [x] Include query-parameters in the `GET /api/v2/task/` endpoint for filtering the task list so that the response only include a subset of matching tasks -> TBD whether this should be done in the backend or in the webclient
- [x] Update PATCH-single-task endpoints
- [x] Add tests
- [x] Test icontains on nullable column

### Expected behavior for PATCH endpoints

* `type`, `modality` and `tags` can be patched by users who have full access to tasks (ref #1803)
* `authors` can only be set upon task creation (either explicitly, for a single-task creation, or through the task-package manifest)

### Side note:

In principle some attributes (e.g. `authors`) may be identical for the tasks of a given taskgroup. For the moment we still choose to have them as single-task attributes, rather than shared task-group attributes. This makes access to these attributes simpler, removing the need for an additional database query.

---

cc @jluethi @lorenzocerrone let me know if this match with our discussion earlier today, or if I forgot something - thanks!

### LINKED ISSUE #1829: Core admin-side `TaskGroup` attributes
The most user-facing new task attributes are defined in #1828, but `TaskV2`s or `TaskGroupV2`s should also include additional metadata that may replace the way we use `task.source`. That is, the first block of the source (all fields which are shared by tasks of the same package) should be transformed into a set of `TaskGroupV2` attributes, so that the only task-specific part is the task name.

Note: for tasks that are added directly (without using task collection), several of the new attributes will be empty - mirroring the fact that their source is currently unstructured. Defining anti-duplication rules for this kind of tasks will require different rules.

---

Based on these new attributes, we'll then be able to build some logic to handle a few specific cases in endpoints.

Some are _existing_ features which will just become smoother:
1. I collect a task and I try to collect it again. The endpoint immediately responds appropriately (e.g. a package with the same metadata already exists). This would be the same behavior as in current main, but without the need to check for the existence of a folder (which is quite useful e.g. in a SSH context).
2. I want collect two independent tasks that have a matching package name (e.g. because they are not canonically normalized). The second API call responds appropriately (e.g. "error: this task collection would use the same folder as another one")

Other are new features related to task-package duplication within the new access-control scheme:
3. I want to collect package X=1.2.3, but the same package&version is already available to me -> the endpoint responds "error: please use the existing package" - ref upcoming issue
4. Consequence of previous point: if I collected package X=1.2.3 privately, then I cannot also collect it a second time and associate it to a usergroup. What I can do is to change the visibility of the existing taskgroup.
5. If I collected package X=1.2.3 privately and then someone else collected it and associated it to a usergroup, then I cannot later make my version "public" - ref upcoming issue.

ychiucco:
closes #1828 
closes #1829 

## Checklist before merging


fractal_server/app/schemas/v2/task.py
```diff
@@ -83,38 +88,55 @@ def validate_commands(cls, values):
         valdictkeys("output_types")
     )
 
+    _category = validator("category", allow_reuse=True)(valstr("category"))
```
tcompa, line 91: allow None - here and below

fractal_server/app/schemas/v2/task.py
```diff
@@ -140,6 +162,15 @@ def val_is_dict(cls, v):
...
```
tcompa, line 165: allow None

fractal_server/app/schemas/v2/task.py
```diff
@@ -83,38 +88,55 @@ def validate_commands(cls, values):
...
+    _modality = validator("modality", allow_reuse=True)(valstr("modality"))
+
+    @validator("tags")
+    def validate_list_of_strings(cls, value):
+        for i, tag in enumerate(value):
+            value[i] = valstr(tag)
```
tcompa, line 97: fix use of valstr

fractal_server/app/routes/api/v2/task.py
```diff
@@ -66,6 +75,25 @@ async def get_list_task(
     if args_schema_non_parallel is False:
         for task in task_list:
             setattr(task, "args_schema_non_parallel", None)
+    if authors is not None:
```
tcompa, line 78: This does work, but let's get back to a single string so that we can include the query in sql

fractal_server/app/routes/api/v2/task.py
```diff
@@ -66,6 +75,25 @@ async def get_list_task(
...
+        task_list = [
+            task
+            for task in task_list
+            if all(
+                author.lower()
+                in (t_author.lower() for t_author in task.authors)
+                for author in authors
+            )
+        ]
+    if tags is not None:
```
tcompa, line 88: Let's postpone implementing this query, so that we can take our time to learn aboute sqlalchemy/sqlite/postgres support for JSON queries.
(please open a new issue)

tests/fixtures_server_v2.py
```diff
@@ -239,6 +239,8 @@ async def __task_factory(
             user_id=user_id,
             user_group_id=user_group_id,
             active=active,
+            origin="other",
+            pkg_name="task_factory_v2",
```
tcompa, line 243: please use task.name here

tests/v2/03_api/test_api_task.py
```diff
@@ -67,6 +77,16 @@ async def test_task_get_list(db, client, task_factory_v2, MockCurrentUser):
...
         assert res.json()[2]["args_schema_parallel"] is None
 
+        # Queries
+        res = await client.get(f"{PREFIX}/?category=CONVERSION")
+        assert len(res.json()) == 2
+        res = await client.get(f"{PREFIX}/?modality=em")
+        assert len(res.json()) == 2
+        res = await client.get(f"{PREFIX}/?category=conversion&modality=em")
+        assert len(res.json()) == 1
+        res = await client.get(f"{PREFIX}/?author=cd e")
```
tcompa, line 87: use %20 for space

fractal_server/tasks/v2/background_operations_ssh.py
```diff
@@ -316,9 +316,19 @@ def background_collect_pip_ssh(
                     python_bin=Path(python_bin),
                 )
 
+                # Prepare some task-group attributes
+                task_group_attrs = dict(
+                    pkg_name=task_pkg.package_name,
+                    version=task_pkg.package_version,
+                )
+                if task_pkg.is_local_package:
+                    task_group_attrs["origin"] = "wheel-file"
```
tcompa, line 325: please open an issue about this missing test

fractal_server/app/routes/api/v2/task.py
ychiucco: https://github.com/fractal-analytics-platform/fractal-server/issues/1863

fractal_server/tasks/v2/background_operations_ssh.py
ychiucco: https://github.com/fractal-analytics-platform/fractal-server/issues/1864

PR #2008, opened 2024-11-06
Update filename and path for task-collection scripts 

### LINKED ISSUE #1995: Change filename for task-collection scripts created from templates
There will be multiple scripts for task collection, task deactivation and task reactivation - all in the same folder.
A reasonable filename structure could be 
```
{task_group.path}/scripts/{timestamp_in_seconds}_{TaskGroupActivity.type}_{name}.sh
```

### LINKED ISSUE #1997: Review structure of `task_group.path` folder
This should be like
```
{task_group.path}
{task_group.path}/venv  # this can be removed during deactivation, and recreated during re-activation
{task_group.path}/scripts
{task_group.path}/{wheel_filename}  # this only applies to task collections based on a user's wheel file
```

Note: this folder should _not_ include any of these:
* temporary files
* logfiles
* JSON files which duplicate some metadata that also exist in the DB

tcompa:
close #1995
close #1997

## Checklist before merging

### DIFF
--- a/tests/v2/03_api/test_api_task_collection_ssh.py
+++ b/tests/v2/03_api/test_api_task_collection_ssh.py
@@ -49,7 +49,7 @@ async def test_task_collection_ssh_from_pypi(
     # Assign FractalSSH object to app state
     app.state.fractal_ssh_list = fractal_ssh_list
 
-    # Override settins with Python/SSH configurations
+    # Override settings with Python/SSH configurations
     current_py_version_underscore = current_py_version.replace(".", "_")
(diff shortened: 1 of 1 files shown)


fractal_server/tasks/v2/utils_templates.py
```diff
@@ -25,6 +27,10 @@ def customize_template(
     script_data = template_data
     for old_new in replacements:
         script_data = script_data.replace(old_new[0], old_new[1])
+    # Create parent folder if needed
+    parent_dir = Path(script_path).parent
```
mfranzon, line 31: why not just `Path(script_path).mkdir(parents=True, exist_ok=True)`?

PR #2084, opened 2024-11-20
Do not `fork` from multi-threaded processes

### LINKED ISSUE #1957: Do not `fork` from multi-threaded processes
Refs:
* https://discuss.python.org/t/concerns-regarding-deprecation-of-fork-with-alive-threads/33555
* https://docs.python.org/3/library/os.html#os.fork
* https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor

> Changed in version 3.12: On POSIX systems, if your application has multiple threads and the [multiprocessing](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing) context uses the "fork" start method: The [os.fork()](https://docs.python.org/3/library/os.html#os.fork) function called internally to spawn workers may raise a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning). Pass a mp_context configured to use a different start method. See the [os.fork()](https://docs.python.org/3/library/os.html#os.fork) documentation for further explanation.
> (https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor)

> Changed in version 3.12: If Python is able to detect that your process has multiple threads, [os.fork()](https://docs.python.org/3/library/os.html#os.fork) now raises a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning).
> We chose to surface this as a warning, when detectable, to better inform developers of a design problem that the POSIX platform specifically notes as not supported. Even in code that appears to work, it has never been safe to mix threading with [os.fork()](https://docs.python.org/3/library/os.html#os.fork) on POSIX platforms. The CPython runtime itself has always made API calls that are not safe for use in the child process when threads existed in the parent (such as malloc and free).
> Users of macOS or users of libc or malloc implementations other than those typically found in glibc to date are among those already more likely to experience deadlocks running such code.
> See [this discussion on fork being incompatible with threads](https://discuss.python.org/t/33555) for technical details of why we’re surfacing this longstanding platform compatibility problem to developers.
>
> (https://docs.python.org/3/library/os.html#os.fork)

---

To do:
* Set e.g. `mp_context=multiprocessing.get_context("spawn")` and check whether anything breaks.

ychiucco:
closes #1957 

## Checklist before merging


ychiucco:
With [22e5998](https://github.com/fractal-analytics-platform/fractal-server/pull/2084/commits/22e5998a4edcf5351d22752264875b68672c3f92), warnings about `os.fork()` disappeared from python 3.12

PR #2144, opened 2024-12-12
Prevent deactivation of packages with github dependencies

tcompa:
## Checklist before merging


fractal_server/tasks/v2/local/deactivate.py
```diff
@@ -189,6 +189,15 @@ def deactivate_local(
                             "task-group attributes."
                         )
 
+                # Fail if `pip_freeze` includes "github", see
+                # https://github.com/fractal-analytics-platform/fractal-server/issues/2142
+                if "github.com" in task_group.pip_freeze:
```
mfranzon, line 194: should we move the string into a more global variable for example in `fractal_server/tasks/utils.py`?

mfranzon (approved):

PR #2559, opened 2025-05-09
Update install_and_deploy link do developement page

EdytaRz:
There is a broken link on install page that i think could be fixed like that. 
Best.


ychiucco:
Thanks @EdytaRz !

PR #2601, opened 2025-05-20
Propagate IMAGE_STATUS_KEY into runner logic

### LINKED ISSUE #2265: Only process a subset of images based on their status
This question becomes relevant when the criterion to use to filter images is like "images where a given task has not successfully run", which cannot be represented in terms of `attribute_filters`.

For the moment, the naive solution would be to expose an additional parameter for job submission (e.g. `images_to_process = ["/tmp/image/1", "/tmp/image/223"]`), which would then make the runner ignore any value of `attribute_filters` or `type_filters` for the first task.
This would not work when there are more than one task in the job, however.

TBD

mfranzon:
close #2265 

## Checklist before merging

### DIFF
--- a/fractal_server/app/runner/v2/runner.py
+++ b/fractal_server/app/runner/v2/runner.py
@@ -8,6 +8,7 @@ from typing import Literal
 
 from sqlalchemy.orm.attributes import flag_modified
 from sqlmodel import delete
+from sqlmodel import select
 from sqlmodel import update
 
 from ....images import SingleImage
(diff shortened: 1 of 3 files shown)


mfranzon:
TODO:

- move `_enriched_image` into `fractal_server.images.tools.py`

- wrap the image status update into a function
```python
 tmp_images = [
                    _enriched_image(
                        img=img,
                        status=HistoryUnitStatusWithUnset.FAILED,
                    )
                    for img in tmp_images
                ]
                # Write current dataset images into the database.
                db_dataset = db.get(DatasetV2, dataset.id)
                db_dataset.images = tmp_images
                flag_modified(db_dataset, "images")
                db.merge(db_dataset)
                db.commit()
```

PR #2749, opened 2025-09-02
Split ssh job submission

### LINKED ISSUE #2431: Split submission of N jobs into two steps (prepare-N-jobs, submit-N-jobs)
This is not relevant for sudo-slurm, but it is important for ssh-slurm - to avoid transfering N input files/scripts one by one.

### LINKED ISSUE #2754: Drop `pre_submission_commands`
We added this experimental feature
https://github.com/fractal-analytics-platform/fractal-server/blob/e6300ac5b8817ec1d4c653a6bafcdf845c8fad38/CHANGELOG.md?plain=1#L910-L914

The goal was to run `module load gpu` before running `sbatch` to cover usage of GPUs at ScienceCluster, but it turned out that this was not actually required - see https://docs.s3it.uzh.ch/cluster/job_submission/#gpus and https://github.com/fractal-analytics-platform/fractal-server/issues/1565#issuecomment-2247325160.

This feature has remained in the code until now, and it creates some friction in development and testing - e.g. as part of https://github.com/fractal-analytics-platform/fractal-server/pull/2749.

I suggest we fully deprecate it (also since re-including it later would only require a moderate effort).

ychiucco:
closes #2431
closes #2754 

## Checklist before merging

### DIFF
--- a/fractal_server/app/runner/executors/slurm_common/base_slurm_runner.py
+++ b/fractal_server/app/runner/executors/slurm_common/base_slurm_runner.py
@@ -397,7 +397,9 @@ class BaseSlurmRunner(BaseRunner):
         logger.debug("[_prepare_single_slurm_job] END")
         return submit_command
 
-    def _send_many_job_inputs(self, *args, **kwargs) -> None:
+    def _send_many_job_inputs(
+        self, *, workdir_local: Path, workdir_remote: Path
+    ) -> None:
         """
         Placeholder method.
 


fractal_server/ssh/_fabric.py
```diff
@@ -447,6 +447,49 @@ def send_file(
                 ),
             )
```
mfranzon, line 449: add `retry_if_socket_error` decorator

PR #2978, opened 2025-11-12
View logs after job folder is zipped

### LINKED ISSUE #2871: How to view logs after job folder is zipped?
When a job is complete, we always create a zip of its job folder. If the zip is larger than a given threshold, we delete the corresponding job folder. When users download the logs zip, they are served the zip file directly. The current threshold is 1MB:
https://github.com/fractal-analytics-platform/fractal-server/blob/24d0b4789c408d50a63f746a9c6a118543e4b2b4/fractal_server/zip_tools.py#L15

This setup breaks the possibility of getting individual logs when the job is complete (provided that a large-enough folder was created), because the `image-log` or `unit-log` endpoints would try to read each file individually.

TBD: How complex is it to read a single file from a zipped folder? If it is doable (and not terribly slow), we could proceed this way. If not, we need to find a better strategy.

ychiucco:
closes #2871 

## Checklist before merging

### DIFF
--- a/fractal_server/app/routes/api/v2/_aux_functions_history.py
+++ b/fractal_server/app/routes/api/v2/_aux_functions_history.py
@@ -103,8 +103,7 @@ def read_log_file(
             )
             return (
                 f"Logs for task '{task_name}' in dataset "
-                f"{dataset_id} are not available:\n"
-                f"both {logfile=} and {archive_path=} do not exist."
+                f"{dataset_id} are not available."
             )
     except Exception as e:
         logger.error(


fractal_server/app/routes/api/v2/_aux_functions_history.py
```diff
@@ -83,10 +85,27 @@ def read_log_file(
     try:
         with open(logfile) as f:
             return f.read()
-    except Exception as e:
+    except OSError as e1:
```
tcompa: We can use `if Path(logfile).exists() / else` directly, so that we don't end up capturing a broader `OSError` class.

fractal_server/app/routes/api/v2/history.py
```diff
@@ -444,11 +445,23 @@ async def get_image_log(
         db=db,
     )
 
+    history_run = await get_history_run_or_404(
+        history_run_id=history_unit.history_run_id, db=db
+    )
+    check_historyrun_related_to_dataset_and_wftask(
```
tcompa: drop

fractal_server/app/routes/api/v2/history.py
```diff
@@ -444,11 +445,23 @@ async def get_image_log(
         db=db,
     )
 
+    history_run = await get_history_run_or_404(
```
tcompa: Get `job` directly

PR #2999, opened 2025-11-18
Project sharing

### LINKED ISSUE #2997: Project-sharing core
# Project sharing

## Data structure

In the current database, users and projects are related through a many-to-many link table `LinkUserProjectV2`, which we have never used for linking a project to more than a single user.

In the new version, this link table gets three new attributes:
* `is_owner: bool`
* `is_verified: bool`
* `permissions: Literal["r", "rw", "rwx"]` (actually a unconstrainted string, in the db schema).

> Implementation detail 1: It could be useful to represent r/rw/rwx as an integer in `[1, 2, 3]` - so that comparisons can be made as `can_read = permissions >= 1`, `can_write = permissions >= 2`, `can_execute = permissions == 3` (possibly using an https://docs.python.org/3/library/enum.html#enum.IntEnum for values). This should be reviewed _after_ the r/rw/rwx implementation is in-place, because it's likely that it is not even needed.

> Implementation detail 2: For the moment we should _not_ explore using 0,1,2,3,4 for non-verified/can_read/can_write/can_execute/is_owner, since the added complexity is likely not justified.

Some additional database constraints are in-place, to make sure we don't represent invalid data:
* A single owner per project (`UNIQUE` constraint on `is_owner`).
* Owner is alwasy verified (`CHECK not (is_owner and not is_verified)`).
* Owner always have full permissions (`CHECK not (is_owner and permissions!="rwx")`).
* Permissions are valid (something like `CHECK permissions in {"r", "rw", "rwx"}`).
   
Examples of valid `LinkUserProjectV2` rows are
```python
# A project I own:
(my_user_id, project_id, is_owner=True, is_verified=True, permissions="rwx")

# After I share a project with user B (and I pick the value of "permissions"), new row like
(user_B_id, project_id, is_owner=False, is_verified=False, permissions="rw")

# After user B accepts the invitation, the row is updated with `is_verified=True` - as in
(user_B_id, project_id, is_owner=False, is_verified=True, permissions="rw")
```

## Project-sharing flow

* User 1 creates a project.
    * DB: A user-1/project link is established, with `is_owner=True`.
* User 1 shares the project with user 2, which they identify through their email `user2@example.org` (*).
    * API: User 1 is free to choose a `permissions` value for this project-sharing invitation, in the API request.
    * API: The response discloses information about whether such user exists (failing if it is not found).
    * DB: A non-verified user-2/project link is established (with `is_owner=False` and `is_verified=False`).
* User 2 now can see a notification about a pending project-sharing invitation (notification system is described elsewhere).
    * API: The notification includes the email of the project owner.
* User 2 rejects the invitation
    * DB: The user-2/project link is deleted.
* User 2 accepts the invitation
    * DB: The non-verified user-2/project link is marked as verified.
    * API: User 2 can always see the email address of the project owner - as long as user 2 has access to the project.
    * API: At any point in the future, user 1 can revoke user-2 access to the project.
    * API: At any point in the future, user 2 can "unsubscribe" from the project.
* API: Any user can always see a list of their owned projects.
* API: Any user can always see a list of projects shared with them that they have access to.
    * API: For each project, 
* API: For each one a user has access to, they can see what permission set they have.
* API: Any user can always see a list of pending project-sharing invitations.

(*) Bulk invites of N users can be included later.

## Project-sharing API

> Note: The specific API is an implementation detail, which can change during development.
```
# ANY USER

# List my owned projects
GET /api/v2/project/
GET /api/v2/project/?is_owner=true
[is_owner=true is the default]

# List "shared with me" projects (note: `is_verified=True` is implied)
GET /api/v2/project/?is_owner=false

# Note: the "list" endpoints do not include rich information like `owner_email` or `permissions` for all their entries. This info is available at a single-project level:
GET /api/v2/project/{project_id}/access-info/
	owner_email = a@a.a
	is_owner = false
	permissions = rw


# OWNER

## See users associated to project
GET /api/v2/project/{project_id}/link/
[

 ("a1@a.a", true, "rw"),
 ("a2@a.a", false, "rw"),
 ("a3@a.a", true, "rwx"),
]
[note: does not include owner]
 
## Share a project
POST /api/v2/project/{project_id}/link/?email=a1@a.a
	permission = r,rw,rwx
[options: 201, 404 (<- one_or_none), 422 (link already exists, verified or not)]

## Change permissions for an existing link
PATCH /api/v2/project/{project_id}/link/?email=a1@a.a
	permissions = ...
[options: 200, 404 (including if user is owner - thanks to a where-is-not-owner]

## Kick out a guest from the project
DELETE /api/v2/project/{project_id}/link/?email=a1@a.a
[options: 204, 404 (including if user is owner - thanks to a where-is-not-owner]

# GUEST

## See current invitations
GET /api/v2/project/invitation/
	[
	("project A", 1, "owner@example.org", "rw"),
	("project B", 2, "owner@example.org", "r"),
	]

## Accept invite
PATCH /api/v2/project/{project_id}/guest-link/
	is_verified = True
[options: 200, 404]

## Reject invite or unsubscribe from project
DELETE /api/v2/project/{project_id}/guest-link/
[options: 204, 404, 422 (e.g. I am the owner)]



# ADMIN


# Admin: list all links, with optional query parameters
GET /admin/v2/linkuserproject/
GET /admin/v2/linkuserproject/?project_id=123 [optional]
GET /admin/v2/linkuserproject/?user_id=123 [optional]
GET /admin/v2/linkuserproject/?is_verified=false [optional]
GET /admin/v2/linkuserproject/?is_owner=true [optional]
[with pagination, order by user_id]

Example of response
	[
	(user_id, project_id, is_verified, is_owner, permissions, user_email, project_name)
	]
```

### LINKED ISSUE #3006: Project-sharing access control
Branching from #2997.

A project owner maintains full access to the same `/api/v2/project/{project_id}/` endpoints that were already in-place in previous versions, while a project guest has limited access.

* Seeing any object in the project tree (project, workflows, datasets, jobs, history objects) requires a link with `is_verified=True`. That is, the guest must have already accepted the invitation.
* Editing the database representation of a project tree (including creating workflows, adding tasks, deleting workflows, deleting datasets, ...) requires both `is_verified=True` and `permissions in ["rw", "rwx"]`.
    * The only exception is that `DELETE /api/v2/project/{project_id}/` is restricted to the owner.
* Running (or stopping) a job for a given project requires both `is_verified=True` and `permissions="rwx"`.
    * Note that this may still fail, if the project belongs to computational resource R1 but the current user is associated to computational resource R2.

Note that the following endpoints have some slightly different access-control rules:
* `/api/v2/project/{project_id}/guest/` endpoints are for the owner only
* `/api/v2/project/{project_id}/access/` endpoints are typically for guests only (the `GET` one is also open to the owner, but it's not a relevant use case)

### LINKED ISSUE #3013: Test access to jobs in case of shared rwx project
Question: does `GET /api/v2/project/{project_id}/job/` list all jobs that ran for the current project (that is, both when submitted by me or by others)?
Current answer: Yes.

Let's add this to a test.

### LINKED ISSUE #3017: Test project-sharing access control
* User 1 creates the project and shares it with user 2 with permissions R
* User 2 can read the project but not patch it
* User 1 edits permissions into RW
* User 2 can read and patch the project but not delete it

### LINKED ISSUE #3022: Fix subquery when user owns more than one project
We observed a `psycopg.errors.CardinalityViolation` error for the subquery https://github.com/fractal-analytics-platform/fractal-server/issues/3005, in a case where
```sql
SELECT projectv2.id 
FROM user_oauth JOIN linkuserprojectv2 ON user_oauth.id = linkuserprojectv2.user_id, projectv2
WHERE linkuserprojectv2.is_owner IS true AND linkuserprojectv2.project_id = projectv2.id;
 id 
----
  6
  7
  8
(3 rows)
```

ychiucco:
closes #2997 
closes #3006
closes #3017
closes #3013
closes #3022 

ref #3012

## Checklist before merging
- [ ] Review all docstrings once again

TODO:
- [x] Verify that constraints are part of database, and not just ORM (by trying with psql)
- [x] Make sure that CHECK constraints are listed in the migration script
- [x] docstrings


fractal_server/app/routes/api/v2/project_share.py
```diff
@@ -0,0 +1,305 @@
...
+    user: UserOAuth = Depends(current_user_act_ver_prof),
+    db: AsyncSession = Depends(get_async_db),
+):
+    # Check current user is project owner
+    await check_user_is_project_owner(
+        user_id=user.id, project_id=project_id, db=db
+    )
+
+    # Get the ID of the linked user
+    linked_user_id = await get_user_id_from_email(user_email=email, db=db)
```
tcompa: What about using the "guest" keyword (here and in other endpoints)? (e.g. this one could be `guest_user_id`)

fractal_server/app/models/linkuserproject.py
```diff
@@ -11,3 +14,28 @@ class LinkUserProjectV2(SQLModel, table=True):
...
+            name="ck_linkuserprojectv2_owner_is_verified",
+        ),
+        CheckConstraint(
+            "NOT (is_owner AND permissions <> 'rwx')",
+            name="ck_linkuserprojectv2_owner_full_permissions",
+        ),
+        CheckConstraint(
+            "permissions IN ('r', 'rw', 'rwx')",
+            name="ck_linkuserprojectv2_valid_permissions",
+        ),
```
tcompa: Remove `ck_{table_name}` from name

fractal_server/app/schemas/v2/sharing.py
```diff
@@ -0,0 +1,43 @@
...
+    user_email: EmailStr
+    is_verified: bool
+    permissions: ProjectPermissions
+
+
+class ProjectShareReadGuest(BaseModel):
+    project_name: str
+    project_id: int
+    owner_email: EmailStr
+    permissions: ProjectPermissions
```
tcompa: Do not validate strings?

tests/v2/test_03_api/test_api_sharing.py
```diff
@@ -0,0 +1,266 @@
...
+            f"/api/v2/project/{project1_id}/link/?email={user4.email}",
+        )
+        assert res.status_code == 204
+
+        # Get list of all users linked to Project1
+        res = await client.get(
+            f"/api/v2/project/{project1_id}/link/",
+        )
+        assert res.status_code == 200
+        assert res.json() == []
```
tcompa: Add block with tests of failures

fractal_server/app/routes/api/v2/project.py
```diff
@@ -73,7 +77,13 @@ async def create_project(
     db.add(db_project)
     await db.flush()
 
-    link = LinkUserProjectV2(project_id=db_project.id, user_id=user.id)
+    link = LinkUserProjectV2(
+        project_id=db_project.id,
+        user_id=user.id,
+        is_owner=True,
+        is_verified=True,
+        permissions=ProjectPermissions.EXECUTE,
```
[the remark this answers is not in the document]
ychiucco:
```sh
$ git grep -h 'class.*StrEnum'
class HistoryUnitStatus(StrEnum):
class HistoryUnitStatusWithUnset(StrEnum):
class JobStatusTypeV2(StrEnum):
class ResourceType(StrEnum):
class ProjectPermissions(StrEnum):
class WorkflowTaskStatusTypeV2(StrEnum):
class TaskType(StrEnum):
class TaskGroupV2OriginEnum(StrEnum):
class TaskGroupActivityStatusV2(StrEnum):
class TaskGroupActivityActionV2(StrEnum):
class DataAuthScheme(StrEnum):
class SubmitTaskType(StrEnum):
class MultisubmitTaskType(StrEnum):
class FakeEnum(StrEnum):
```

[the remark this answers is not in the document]
tcompa: Great - thanks for checking.
Then what should be done (elsewhere) is rather renaming `TaskGroupV2OriginEnum` into `TaskGroupV2Origin`

Issue #3040, opened 2025-12-01
Add database constraint for jobs (through partial index)


tcompa:
We currently have an ednpoint-level check like
```python
    # Check that no other job with the same dataset_id is SUBMITTED
    stm = (
        select(JobV2)
        .where(JobV2.dataset_id == dataset_id)
        .where(JobV2.status == JobStatusType.SUBMITTED)
    )
    res = await db.execute(stm)
    if res.scalars().all():
        raise HTTPException(
            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
            detail=(
                f"Dataset {dataset_id} is already in use in submitted job(s)."
            ),
        )
```
but we observed at least once a situation where two "submitted" jobs co-existed.

---

It'd be best to make this more robust, by also having a postgres unique index on "jobs with dataset_id = X and status = submitted (where X is not null)".
I think a partial index should be appropriate:
* https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#partial-indexes
* "Example 11.3. Setting up a Partial Unique Index" section of https://www.postgresql.org/docs/current/indexes-partial.html

If we are able to implement this at db level, we could:
* leave the current check in-place, to provide a user-friendly error message
* rely on the db check to avoid inconsistent state in some unexpected scenarios (this would result in a 500 response and `IntegrityError` internal error - which is anyway much better than having two submitted jobs)

PR #3044, opened 2025-12-01
Max one submitted Job per Dataset

ychiucco:
closes https://github.com/fractal-analytics-platform/fractal-server/issues/3040

## Checklist before merging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 > NOTE: This version requires running a data-migration script (`fractalctl update-db-data`).
 
+> WARNING: Before upgrading to this version, make sure that no jobs are marked as submitted in the current database tables.
+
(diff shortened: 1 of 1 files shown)


tests/v2/test_02_models/test_model_job.py
```diff
@@ -0,0 +1,89 @@
...
+        user_email="user@example.org",
+        dataset_dump={},
+        workflow_dump={},
+        project_dump={},
+        first_task_index=0,
+        last_task_index=0,
+        attribute_filters={},
+        type_filters={},
+    )
+
```
tcompa: Let's add another case:
```
job1: dataset_id = null, status = submitted
job2: dataset_id = null, status = submitted
```
what is the outcome?

(to be clear: this situation should never take place, but it's best to be aware of what it would lead to)

PR #3191, opened 2026-02-16
Workflow templates

### LINKED ISSUE #3186: Workflow templates - core
This is the core issue for the new workflow-template feature. Specific additions will be discussed elsewhere.

## Preliminary info

We'd rather never send users' IDs via the API for standard users, but only their emails (obtained via `JOIN`s).

## Data structure

New `workflow_template` table:

* id  
* Three columns that make a unique set (guaranteed via a unique constraint):  
  * user_id: non-nullable foreign key for the template owner  
  * name: non-nullable string (with the usual API validation logics)  
  * version: **non-nullable** (positive) integer  
* Creation timestamp (use naming similar to other tables)
* `user_group_id`: nullable foreign key with ondelete “set null” option, for read access to the template  
  * NOTE: There isn't any non-duplication constraint on the tuple `(user_group_id, name, version)`. That is, we accept coexistence of  
    * `user1@example.org / my-template-name / version=1 / shared with All`
    * `user2@example.org / my-template-name / version=1 / shared with All`
* `description`: optional non-empty string  
* `data`: non-nullable JSON column storing the equivalent of a workflow-export response
    * Note: it has to be JSON, not JSONB

Updates to `workflowv2` table:

* A new `template_id` nullable foreign key pointing to `workflow_template`, with ondelete-set-null for when the template is deleted.

### All users:

- [x]  `GET /api/v2/workflow_template/` gives a list of available templates  
   * Returns a list of `WorkflowTemplateRead` objects, without stripping anything.
   * Query parameter `is_owner: bool = False` restricts listing to my own templates.
   * Filtering query parameters:
      * `user_email`, exact match
      * `name`, via `icontains`
      * `version`, exact match
   * Two sorting options `sort_by: Literal['user-name-version', 'timestamp'] = 'user-name-version`
   * With pagination
- [x] `GET /api/v2/workflow_template/{template_id}/` is the standard GET endpoint - unclear whether it's needed, but easy to implement

### Template producer (access control: template-owner only)

- [x] `POST /api/v2/workflow_template/?workflow_id=123` creates a workflow template based on an existing workflow and additional metadata.
   * The request body includes `name`, `version`, `user_group_id`, `description`.
   * `user_id` is inferred by who makes the API call  
   * `user_group_id` must be for a group that current user belongs to
   * Non-duplication of `(user_id, name, version)` is checked manually, rather than by catching an integrity error.
   * The JSON data is obtained by wrapping the existing export-workflow endpoint. This requires a refactor of the export-workflow endpoint first, or a small code duplication.
- [x] `GET /api/v2/workflow_template/{template_id}/export/` is the complementary endpoint for the import one below. It strips all properties which are not needed for imports and which are instance-specific: all IDs, user email, creation timestamp, 
- [x] `POST /api/v2/workflow_template/import/?user_group_id=123` creates a workflow template based on a full `WorkflowTemplateCreate`-compliant JSON object (provided in the request body).
   * `user_id` is inferred by who makes the API call
   * `user_group_id` is an optional query parameter, to be selected among the groups that the current user belongs to
   * The non-duplication constraint still apply, the same as for a standard POST. This would lead to a 422, and the user can re-submit (e.g. after bumping the version from `N` to `N+1`).
- [x] `PATCH /api/v2/workflow_template/{template_id}/` lets the template owner edit description or `user_group_id`
    * `user_group_id` must be a group that the current user belongs to
- [x] `DELETE /api/v2/workflow_template/{template_id}/` lets the owner delete a template.  
   * This has no constraint on whether someone is “using” this template. It will not fail.  
  
### Template consumer (access control: owner or users of authorized groups)
  
- [x] `POST /api/v2/project/{project_id}/workflow/import-from-template/?template_id=123` is the equivalent of import-workflow, but it uses a template as a source.
   * This requires a refactor of the core import-workflow feature, so that we can re-use it in both endpoints.
   * This still includes the `HTTPExceptionWithData` kind of failure.
   * This also sets `workflow.template_id` to `template_id`

### LINKED ISSUE #3198: Forbid `WorkflowTemplate` with empty `task_list`

### LINKED ISSUE #3213: Add fractal-server version to workflow-template schema
It is not necessary that we show it in the fractal-web UI, and for the moment it should not determine any specific behavior. But we should keep track of it in the database, to be able to have version-specific behaviors in the future.

### LINKED ISSUE #3221: In flexible-import version, version 0.2.2 is considered more recent than 0.2.14

ychiucco:
closes #3186
closes #3213
closes #3198
closes #3221

- [x] Re-include extra=forbid in dump model
- [x] ~~Review whether indices should be added~~ moved to https://github.com/fractal-analytics-platform/fractal-server/issues/3203
- [x] Make sure that user-facing API-error messages do not include `user_id` or `user_group_id`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,13 @@
 # 2.20.0 (unreleased)
 
 * API:
-    * Introduce version flexibility in workflow-import endpoint (\#3154).
+    * Introduce `/api/v2/workflow-template/` API block (\#3191).
+    * Introduce `POST /project/{project_id}/workflow/import-from-template/` endpoint (\#3191).
+    * Introduce flexible version matching in workflow-import endpoint (\#3154).
     * Handle internal error with `_write_shutdown_file` in `/job/{job_id}/stop/` endpoints (\#3214).
     * Handle `TypeError` inside images `aggregate_attributes` method (\#3223, \#3226).
 * Database:
+    * Introduce new `workflow_template` table and `WorkflowV2.template_id` (\#3191).
     * Enforce non-duplication constraints for `TaskGroupV2` and `TaskGroupActivityV2` via `unique` database checks (\#3185).
 * Application:
     * Improve lifespan logging (\#3201).


fractal_server/app/schemas/v2/dumps.py
```diff
@@ -58,6 +58,7 @@ class WorkflowDump(BaseModel):
     id: int
     name: str
     project_id: int
+    template_id: int | None = None
```
tcompa: We should not propagate workflow-origin information in the dump object, which is rather meant to store a copy of the workflow at job-submission time and to provide basic information about the job when the corresponding workflow is deleted.

fractal_server/app/schemas/v2/dumps.py
```diff
@@ -54,7 +54,6 @@ class WorkflowTaskDump(BaseModel):
 
 
 class WorkflowDump(BaseModel):
-    model_config = ConfigDict(extra="forbid")
```
tcompa: I suggest we rather exclude extra fields from the `dump` call in the job-submission endpoint, and keep this schema restrictive.

fractal_server/app/routes/api/v2/_aux_functions.py
```diff
@@ -568,3 +570,64 @@ async def _get_user_resource_id(user_id: int, db: AsyncSession) -> int | None:
     )
     resource_id = res.scalar_one_or_none()
     return resource_id
+
+
```
tcompa: Move aux functions to new module

fractal_server/app/routes/api/v2/workflow_template.py
```diff
@@ -0,0 +1,293 @@
...
+@router.get(
+    "/workflow_template/{template_id}/",
+    response_model=WorkflowTemplateRead,
+)
+async def get_workflow_template(
+    template_id: int,
+    user: UserOAuth = Depends(get_api_guest),
+    db: AsyncSession = Depends(get_async_db),
+) -> WorkflowTemplateRead:
+    res = await db.execute(
```
tcompa: Add access control (requires read access)

fractal_server/app/routes/api/v2/workflow_template.py
```diff
@@ -0,0 +1,293 @@
...
+    user: UserOAuth = Depends(get_api_guest),
+    db: AsyncSession = Depends(get_async_db),
+    pagination: PaginationRequest = Depends(get_pagination_params),
+) -> PaginationResponse[WorkflowTemplateRead]:
+    page = pagination.page
+    page_size = pagination.page_size
+
+    stm = select(WorkflowTemplate, UserOAuth.email).join(
+        UserOAuth, UserOAuth.id == WorkflowTemplate.user_id
+    )
```
tcompa: Add access control (owner or group-access)

fractal_server/app/routes/api/v2/workflow_template.py
```diff
@@ -0,0 +1,293 @@
...
+        name=template_create.name,
+        version=template_create.version,
+        db=db,
+    )
+    data = await export_workflow(
+        project_id=workflow.project_id,
+        workflow_id=workflow_id,
+        user=user,
+        db=db,
+    )
```
tcompa: Use new aux function rather than endpoint call

fractal_server/app/routes/api/v2/_aux_functions_templates.py
```diff
@@ -0,0 +1,68 @@
...
+                    "You are not authorized to "
+                    f"WorkflowTemplate[{template_id}]."
+                ),
+            )
+    return template
+
+
+async def _check_template_duplication(
+    *, user_id: int, name: str, version: int, db: AsyncSession
+):
```
tcompa: docstring

fractal_server/app/routes/api/v2/workflow_import.py
```diff
@@ -211,7 +212,47 @@ async def import_workflow(
...
+    return await _import_workflow(
+        project_id=project_id,
+        template_id=template_id,
+        workflow_import=WorkflowImport(**template.data),
+        user=user,
+        db=db,
+    )
 
+
+async def _import_workflow(
```
tcompa: move up

fractal_server/app/routes/api/v2/workflow_template.py
```diff
@@ -0,0 +1,327 @@
...
+    page = pagination.page
+    page_size = pagination.page_size
+
+    stm = (
+        select(WorkflowTemplate, UserOAuth.email)
+        .join(UserOAuth, UserOAuth.id == WorkflowTemplate.user_id)
+        .where(
+            or_(
+                WorkflowTemplate.user_id == user.id,
+                exists().where(
```
tcompa: https://github.com/fractal-analytics-platform/fractal-server/blob/35794c7b188fd55e51c8f113daa62c5f846d10b7/fractal_server/app/routes/api/v2/task_group.py#L121-L127

fractal_server/app/routes/api/v2/workflow_template.py
```diff
@@ -0,0 +1,327 @@
...
+    total_count = res_total_count.scalar()
+    if page_size is None:
+        page_size = total_count
+    else:
+        stm = stm.offset((page - 1) * page_size).limit(page_size)
+
+    res = await db.execute(stm)
+    templates_and_user_email = res.all()
+
+    return dict(
```
tcompa: This response should also include a list of all available emails. This clashes with the nice pagination feature we have above - maybe we move it to a different small-scope endpoint like `GET /api/v2/workflow_template/available-emails/`?

TBD

fractal_server/app/routes/api/v2/workflow_template.py
```diff
@@ -0,0 +1,327 @@
...
+# ALL USERS endpoints
+
+
+@router.get(
+    "/workflow_template/",
+    response_model=PaginationResponse[WorkflowTemplateRead],
+)
+async def get_workflow_template_list(
+    is_owner: bool = False,
+    user_email: str | None = None,
```
tcompa: add optional `template_id` query param

fractal_server/app/routes/api/v2/_aux_functions_templates.py
```diff
@@ -0,0 +1,137 @@
...
+        .where(WorkflowTemplate.name == name)
+        .where(WorkflowTemplate.version == version)
+    )
+    duplicate = res.one_or_none()
+    if duplicate:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail=(
+                "There is already a WorkflowTemplate with "
+                f"{user_id=}, {name=}, {version=}."
```
tcompa: Rephrase as:
> The current user already own a workflow template with {name=} and {version=}.

In general: user-facing messages should not use user IDs.

fractal_server/app/routes/api/v2/_aux_functions_templates.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+    """
+    res = await db.execute(
+        select(WorkflowTemplate)
+        .where(WorkflowTemplate.user_id == user_id)
+        .where(WorkflowTemplate.name == name)
+        .where(WorkflowTemplate.version == version)
+    )
+    duplicate = res.one_or_none()
```
tcompa: I would replace this with `db.get(WorkflowTemplate, ...)`, to make use of the primary key directly.

PR #3270, opened 2026-04-10
Review usage of `startswith` in queries

### LINKED ISSUE #3263: Review usage of startswith
In

https://github.com/fractal-analytics-platform/fractal-server/blob/4ba3c7610dfebe1a2d0ac77ac9b67dc10bd7b678/fractal_server/app/routes/auth/_aux_auth.py#L235

it would be best to use one of the s3-aware aux functions, but we need to review whether they can be used within a sqlalchemy query.

1. Prepare an isolated example where `url_is_relative_to` is used inside a db query.
2. If 1 works:
    * Replace it in the aux function above
    * Rewrite the db query in the project-ownership-transfer endpoint
3. If 1 doesn't work, review whether the aux function above is robust to edge cases

ychiucco:
ref #3263


ychiucco:
I've tried using [hybrid attributes/methods](https://docs.sqlalchemy.org/en/20/orm/extensions/hybrid.html), but that just shifted the problem to the `.expression` method

PR #3276, opened 2026-04-14
Admin-console users export

### LINKED ISSUE #3269: Admin console: user export
For keeping an overview of users and integrate with the reporting tools, it would be very nice to be able to export users with a bit of extra information directly from Fractal web. Specifically, I want to know:
- Who were the all or active users (in a given period or over the whole server runtime)? Active = have run a workflow, with option to filter for relevant months. All = all registered users
- What are the user project directory/project directories
- What are the user slurm account (if set)
- What groups are users associated to

And that's something I'd like to download as a CSV from the admin console (or via the CLI client, but the admin console is easier for me).

What I currently do is download the jobs csv & list the unique users that have submitted something. And to keep track of group assignment, I keep a separate list (but it's not in sync with the server-side setting => hard to manage).
For project directory & slurm account overviews, we'd done custom scripts so far to collect that info.

tcompa:


fractal_server/app/routes/admin/v2/users_csv.py
```diff
@@ -0,0 +1,118 @@
...
+                UserGroup,
+                LinkUserGroup.group_id == UserGroup.id,
+            )
+        )
+        .where(LinkUserGroup.user_id == UserOAuth.id)
+    )
+
+    stm = (
+        select(
+            Bundle(
```
tcompa, line 76: cc @ychiucco

fractal_server/app/routes/admin/v2/users_csv.py
```diff
@@ -0,0 +1,118 @@
...
+    db: AsyncSession = Depends(get_async_db),
+) -> StreamingResponse:
+    """
+    Provide csv table of users and some of their properties.
+    """
+    stm_num_job = (
+        select(func.count(JobV2.id))
+        .join(ProjectV2, ProjectV2.id == JobV2.project_id)
+        .join(LinkUserProjectV2, LinkUserProjectV2.project_id == ProjectV2.id)
+        .where(LinkUserProjectV2.user_id == UserOAuth.id)
```
ychiucco: missing `.where(LinkUserProjectV2.is_owner.is_(True)`?

fractal_server/__init__.py
```diff
@@ -1 +1 @@
-__VERSION__ = "2.22.2"
+__VERSION__ = "2.22.3a0"
```
ychiucco, line 1: Shouldn't this change be made in the main branch?

ychiucco (changes requested):

fractal_server/__init__.py
tcompa: For stable releases, yes.
For prereleases, we sometimes tag them from within non-`main` branches so that we can quickly test them before merging a PR (as in this case, see 033eb102ddd17e8a5c4d310998d006314a6de60f).

fractal_server/app/routes/admin/v2/users_csv.py
tcompa: Good catch!

The question here is what should count as a user job.

What is currently in-place is wrong, because when a project is shared to users A and B their job numbers would both increase by one.

As a workaround, I'm now changing this into a check on the `JobV2.user_email`. This will work fine, but not in case a user changes email.

tcompa: cc @jluethi 

Because of shared projects, we cannot count jobs-per-user by looking for jobs associated to projects.
We then moved back to a less robust check, which compares `JobV2.user_email` with `UserOAuth.email` to identify jobs by a given user. This works, but not in the case where a user changes email over time.

PR #3294, opened 2026-04-23
Introduce `GET /api/v2/dataset/`

### LINKED ISSUE #3293: Introduce `GET /api/v2/dataset/`
This does not depend on #3217, but once we have tags they will be included in this endpoint as well.

Here is the expected UI. The endpoint specs should be a consequence of this.

---

A paginated table with filtering options on top.
Filtering takes place backend-side.

Filters:
* By project ID (TBD whether a dropdown works). This is made somewhat irrelevant by the fact that I can click on the project name and get to the project page, with all its datasets listed.
* By dataset name (`icontains`)
* By project name (`icontains`)
* By "only owner" (-> only datasets for projects that I own)
* After #3217, filter by tag.

Sorting:
* Default sort is by timestamp created (most recent first)
* [not in scope for this PR] Include different sorting options. This will be defined after #3217.

Columns:
* dataset name -> link to dataset page
* project name -> link to project page
* timestamp_created
* zarr_dir (truncated)
* [_only if postgres has  function for it_] number of images
* After #3217: tags

ychiucco:
#3293


fractal_server/app/routes/api/v2/dataset.py
```diff
@@ -285,3 +292,46 @@ async def import_dataset(
...
+    user: UserOAuth = Depends(get_api_guest),
+    db: AsyncSession = Depends(get_async_db),
+    pagination: PaginationRequest = Depends(get_pagination_params),
+) -> PaginationResponse[DatasetV2]:
+    stm = (
+        select(DatasetV2)
+        .join(LinkUserProjectV2, LinkUserProjectV2.user_id == user.id)
+        .join(ProjectV2, ProjectV2.id == LinkUserProjectV2.project_id)
+        .where(DatasetV2.project_id == ProjectV2.id)
+        .order_by(DatasetV2.timestamp_created)
```
tcompa: Is this "more recent first", or would we need `desc`?

PR #3308, opened 2026-05-05
Store task-group ID in task-group collection folder

### LINKED ISSUE #3300: Store task-group ID in task-group collection folder
While going through a cleanup of the tasks folder on a given instance, we have plenty of "ghost" task-group folders.
Based on their path, we know the user ID, the package name and its version. But we don't have a quick way to find the task-group ID, and in some cases we could not find them in the database.

To do: always add a `fractal_task_group_id.txt` file in `TaskGroupV2.path`, with just the ID.

ychiucco:
closes #3300 

### DIFF
--- a/fractal_server/tasks/utils.py
+++ b/fractal_server/tasks/utils.py
@@ -5,6 +5,8 @@ COLLECTION_LOG_FILENAME = "collection.log"
 COLLECTION_FREEZE_FILENAME = "collection_freeze.txt"
 FORBIDDEN_DEPENDENCY_STRINGS = ["github.com"]
 
+TASK_GROUP_ID_FILENAME = "fractal_task_group_id.txt"
+
 
 def get_log_path(base: Path) -> Path:
     return base / COLLECTION_LOG_FILENAME
--- a/fractal_server/tasks/utils.py
+++ b/fractal_server/tasks/utils.py
@@ -1,8 +1,6 @@
 from pathlib import Path
 
-COLLECTION_FILENAME = "collection.json"
 COLLECTION_LOG_FILENAME = "collection.log"
-COLLECTION_FREEZE_FILENAME = "collection_freeze.txt"
 FORBIDDEN_DEPENDENCY_STRINGS = ["github.com"]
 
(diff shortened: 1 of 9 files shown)


fractal_server/tasks/v2/local/collect.py
```diff
@@ -101,6 +101,14 @@ def collect_local(
                 Path(task_group.path).mkdir(parents=True)
                 logger.info(f"Created {task_group.path}")
 
+                # Write txt file in task_group.path
+                txt_path = (
+                    Path(task_group.path) / "fractal_task_group_id.txt"
```
tcompa: The filename should be hardcoded only once, and then imported where it is needed.

fractal_server/tasks/v2/ssh/collect.py
```diff
@@ -127,6 +127,18 @@ def collect_ssh(
                     fractal_ssh.mkdir(folder=task_group.path, parents=True)
                     fractal_ssh.mkdir(folder=script_dir_remote, parents=True)
 
+                    # Write txt file locally and send it to remote path
```
tcompa: I think we have a remote-write ssh function/method

tcompa (changes requested):

fractal_server/tasks/utils.py
```diff
@@ -5,6 +5,8 @@
 COLLECTION_FREEZE_FILENAME = "collection_freeze.txt"
 FORBIDDEN_DEPENDENCY_STRINGS = ["github.com"]
 
+TASK_GROUP_ID_FILENAME = "fractal_task_group_id.txt"
```
tcompa, line 6: Since we are already looking at this file, could you double check which ones of the variables defined above are actually in use?

ychiucco: I've remove `COLLECTION_FILENAME` and `COLLECTION_FREEZE_FILENAME`.

`COLLECTION_LOG_FILENAME` is only used inside the `get_log_path` function, which is then used many times elsewhere.

tcompa (approved):

PR #3314, opened 2026-05-11
Let admins re collect a task group from scratch

### LINKED ISSUE #3312: Let admins re-collect a task-group from scratch
1. A specific env variable (e.g. `FRACTAL_ENABLE_TASK_GROUP_RESET`, false by default) enables this endpoint. This is because of the risks associated with it (see below), and because it's only meant to be used as part of an instance migration where we want to maintain some backwards compatibility for existing tasks.
2. This only applies to non-active task groups. If a task group is active, the endpoint fails with a 422 response.
3. The background task will use the task-group properties to restart the original collection, ignoring the existing `env_info` property (which would be the pip-freeze or pixi-lock data).
    * **NOTE**: For this reason, this is a **non-reversible** action. We loose track of which dependencies were used in the past for this specific task group. This is why this should not be enabled by default.
4. We may decide what to do with the original pinned dependencies (if any). I think they should be discarded by default, because they were very specific for the old cluster - or there should at least be an option to ignore them. TBD.

The usage would be as follows:
1. If a task is already active and it works, nothing to do.
2. If a task is already active but it does not work, the user should deactivate it and try to reactivate it.
3. If a task is not active, the user should try to reactivate.
4. If the reactivation in 2 or 3 fails, the admin can proceed with a re-collection.

tcompa:


fractal_server/app/routes/admin/v2/task_group_reset.py
```diff
@@ -0,0 +1,178 @@
...
+        reset_local,
+        task_group_id=task_group.id,
+        task_group_activity_id=task_group_activity.id,
+        resource=resource,
+        profile=profile,
+        wheel_file=None,
+    )
+
+    logger.debug(
+        "Task-collection endpoint: start background collection "
```
tcompa: Clean up logs

fractal_server/app/routes/admin/v2/task_group_reset.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+    logger.info(
+        f"Replacing {task_group.pinned_package_versions_post=} with "
+        f"{req_body.pinned_package_versions_post=}"
+    )
+    task_group.pinned_package_versions_pre = (
+        req_body.pinned_package_versions_post
+    )
+    db.add(task_group)
+    await db.commit()
```
tcompa, line 196: TBD what if something fails later?

tcompa: Also: remove venv info

fractal_server/app/routes/admin/v2/task_group_reset.py
```diff
@@ -0,0 +1,178 @@
...
+    await db.refresh(task_group_activity)
+
+    background_tasks.add_task(
+        reset_local,
+        task_group_id=task_group.id,
+        task_group_activity_id=task_group_activity.id,
+        resource=resource,
+        profile=profile,
+        wheel_file=None,
```
tcompa: Remove wheel_file

fractal_server/tasks/v2/local/reset.py
```diff
@@ -0,0 +1,190 @@
...
+from ._utils import rmtree_nofail
+
+
+def reset_local(
+    *,
+    task_group_activity_id: int,
+    task_group_id: int,
+    resource: Resource,
+    profile: Profile,
+    wheel_file: FractalUploadedFile | None = None,
```
tcompa: Remove.

fractal_server/tasks/v2/local/reset.py
```diff
@@ -0,0 +1,190 @@
...
+    NOTE:  since this function is sync, it runs within a thread - due to
+    starlette/fastapi handling of background tasks (see
+    https://github.com/encode/starlette/blob/master/starlette/background.py).
+
+
+    Args:
+        task_group_id:
+        task_group_activity_id:
+        resource: Resource
+        wheel_file:
```
tcompa: Remove

fractal_server/tasks/v2/local/reset.py
```diff
@@ -0,0 +1,190 @@
...
+                error_msg = f"{task_group.path} does not exist."
+                logger.error(error_msg)
+                fail_and_cleanup(
+                    task_group=task_group,
+                    task_group_activity=activity,
+                    logger_name=LOGGER_NAME,
+                    log_file_path=log_file_path,
+                    exception=FileNotFoundError(error_msg),
+                    db=db,
+                )
```
tcompa: Remove - handle with the generic error handling.

fractal_server/tasks/v2/local/reset.py
```diff
@@ -0,0 +1,190 @@
...
+                # Update task_group data
+                logger.info("Update TaskGroupV2 - start")
+                task_group.env_info = pip_freeze_stdout
+                task_group.active = True
+                task_group = add_commit_refresh(obj=task_group, db=db)
+                logger.info("Update TaskGroupV2 - end")
+
+                # Finalize (write metadata to DB)
+                logger.info("finalising - START")
+                task_group.active = True
```
tcompa: remove

fractal_server/app/routes/admin/v2/task_group_reset.py
```diff
@@ -0,0 +1,178 @@
...
+        f"{req_body.pinned_package_versions_pre=}"
+    )
+    task_group.pinned_package_versions_pre = (
+        req_body.pinned_package_versions_pre
+    )
+    logger.info(
+        f"Replacing {task_group.pinned_package_versions_post=} with "
+        f"{req_body.pinned_package_versions_post=}"
+    )
+    task_group.pinned_package_versions_pre = (
```
ychiucco: `_post`

ychiucco (approved):

fractal_server/app/routes/admin/v2/task_group_reset.py
tcompa: Thanks!

8035f18960

Issue #1507, opened 2024-05-17
Reproduce/understand timeout errors


tcompa:
Ref:
* https://github.com/fractal-analytics-platform/fractal-server/issues/1463#issuecomment-2106919895
* https://github.com/fractal-analytics-platform/fractal-server/issues/1154

I'll summarize here a deep dive we had with @mfranzon, with the simplest reproducible example we could find. We don't have a full understanding yet, but I think we are reproducing the actual behavior.

1. Gunicorn has a "master" process (the "arbiter", which starts one process per worker.
2. Arbiter and workers have a heartbeat system in place, so that arbiter can verify that workers are alive.
3. This arbiter-side check takes place every `timeout` interval. Whenever the arbiter checks, if the latest update of a worker status took place more than a `timeout` interval ago, the worker receives a SIGABRT signal (and a timeout error).
4. Workers are initialized with a timeout set to half of `timeout` (see https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L589-L591), and they are expected to update their status every such interval.
5. The docs about timeout (https://docs.gunicorn.org/en/stable/settings.html#timeout) is confusing, probably because it was first written for synchronous workers (not our case), where the `timeout` parameter also serves as a request-time limit. For our workers, this is not the case. Here are some clarifications: https://github.com/benoitc/gunicorn/issues/3129 (see for instance "Gthread-workers may process a single request indefinitely, even if --timeout is set.").
6. The status update (by the worker) corresponds to updating the last-modified time of a temporary file. The status check then reads this information.
7. A frequent issue (especially when running in containers, or debian-based Linux distributions) is that the default working directory for these temporary files is `/tmp` and it is not mounted as a `tmpfs` filesystem. In some cases (e.g. AWS EBS volumes), this may introduce huge slowdowns related to the status-update operation. We don't think this is relevant in our case (as we briefly checked by setting `--worker-tmp-dir` to `/run`, which in Ubuntu is mounted as `tmpfs`). Relevant Gunicorn FAQ: https://docs.gunicorn.org/en/22.0.0/faq.html#how-do-i-avoid-gunicorn-excessively-blocking-in-os-fchmod.

With this better understanding, we now know that a timeout error (when using async workers, as we do) means that the worker was "late" in updating its status, and then the arbiter killed it. This does *not* immediately mean that the request took long (and that's why https://github.com/fractal-analytics-platform/fractal-server/issues/1490 was not an actual issue).

[...]

Issue #1499, opened 2024-05-15
Catch SIGABRT


tcompa:
EDIT we now observed that SIGABRT does not lead to execution of the shutdown lifespan logic. If we can get this to work, then @mfranzon explored a promising route for cleaning up "orphan" jobs - ref https://github.com/fractal-analytics-platform/fractal-server/pull/1506).

----

(requires #1498)

1. Verify what is needed upon shutdown (e.g. should we call some cleanup function for the db engines?).
2. Verify which signals trigger execution of the shutdown (i.e. post-`yield` operations).

We are especially interested in the following situation
* Start fractal-server through gunicorn (with a single worker)
* ~~Trigger a timeout error (with a fake endpoint that takes long)~~ or directly send an abort signal to the correct pid
* Is the shutdown logic executed?

If the answer is yes, we should add the following:
* Query the DB for all V1 and V2 jobs that are marked as submitted
* Mark all these jobs as failed
* Append a line to these jobs logs, saying something like "Failed because the app was shut down"
* If possible, also update the dataset history.

tcompa:
Note: it turns out that in the current version the sigabrt does not trigger the lifespan shutdown. This can probably be changed in fastapi or uvicorn - TBD

tcompa:
> Note: it turns out that in the current version the sigabrt does not trigger the lifespan shutdown. This can probably be changed in fastapi or uvicorn - TBD

This changes if we use a custom UvicornWorker class and subclass the `init_signals` and `handle_abort` method - see proof of principle in #1506, where we are converting SIGABRTs into SIGTERMs.

The nice aspect of this solution is that it is opt-in:
* The standard `fractalctl start` would not include this feature
* For `gunicorn`, this feature would only be enabled when we modify the startup command (changing the `--worker-class` option so that it points to this new class).

This means that if we ever hit an issue due to this change of signals, we can roll-back the feature by restarting `gunicorn` and using the standard uvicorn worker. That's a useful safety net, since this is a bit of an advanced feature which we cannot fully test programmatically.

tcompa:
See https://github.com/fractal-analytics-platform/fractal-server/pull/1506#issuecomment-2124111370 for a minor improvement of the current proof of principle

PR #3343, opened 2026-05-22
Add `LOG_CONFIG_FILE` environment variable use config file for logging

dominique-roux:
Co-authored-by: Copilot <copilot@github.com>

```bash
```


tcompa:
cc @dominique-roux is there a reason for not including https://github.com/fractal-analytics-platform/fractal-server/blob/main/fractal_server/gunicorn_fractal.py in this PR?

Note that the relevance of that worker wrapper should be reviewed (independently on this PR):
* It was introduced as part of a [broader analysis on intermittent worker timeouts](https://github.com/fractal-analytics-platform/fractal-server/issues/1507)
* [While handling it, we observed that SIGABRT was not handled correctly](https://github.com/fractal-analytics-platform/fractal-server/issues/1499).
* The [upstream issue](https://github.com/Kludex/uvicorn-worker/issues/32) still exists, but it's unclear to me whether the conditions that made it relevant for our deployments still exist. We made quite some related progress over time (notably moving job execution to the one-thread-per-Fractal-job pattern, reducing/removing the use of blocking operations in the API endpoints, and writing Gunicorn heartbeat file to the appropriate folders), and I suspect that we could do without the wrapper.

tcompa:
> Note that the relevance of that worker wrapper should be reviewed (independently on this PR):

To be more precise: we should review the relevance of `FractalWorker`, while `FractalGunicornLogger` is a narrowly-scoped and reasonable feature - supported as in https://gunicorn.org/reference/settings/#logger_class

tcompa:
cc @dominique-roux
I committed https://github.com/fractal-analytics-platform/fractal-server/pull/3343/commits/dfad2cdc7cbd1a25d7d0230cad9cfe133467ecce and then reverted it - so that we can discuss it.

Would this one be a viable solution?

The advantage is that we remove one of the "global" variables (which is not a great pattern, in my opinion), and we also remove the block handling `_CONFIG_ERROR` from `set_logger` (which is not the right place for logging information about startup errors).

Unrelated: it's not clear to me how the current implementation works. The comment says that the log is emitted only once, but what prevents emitting it again every time `set_logger` is called?

tcompa:
> cc @dominique-roux I committed [dfad2cd](https://github.com/fractal-analytics-platform/fractal-server/commit/dfad2cdc7cbd1a25d7d0230cad9cfe133467ecce) and then reverted it - so that we can discuss it.
> 
> Would this one be a viable solution?
> 
> The advantage is that we remove one of the "global" variables (which is not a great pattern, in my opinion), and we also remove the block handling `_CONFIG_ERROR` from `set_logger` (which is not the right place for logging information about startup errors).
> 
> Unrelated: it's not clear to me how the current implementation works. The comment says that the log is emitted only once, but what prevents emitting it again every time `set_logger` is called?

If you agree with this change, I'll then update the corresponding tests.

fractal_server/config/_main.py
```diff
@@ -27,6 +27,8 @@ class Settings(BaseSettings):
             Logging-level threshold for logging
             Only logs of with this level (or higher) will appear in the console
             logs.
+        FRACTAL_LOG_CONFIG_FILE:
+            FIXME
```
tcompa: Self-reminder: add description

docs/internals/logging.md
```diff
@@ -10,7 +10,7 @@ documentation](../code_reference/logger.md).
 
 ## External config file
 
-Set the `LOG_CONFIG_FILE` environment variable to the path of a YAML
+Set the `FRACTAL_LOG_CONFIG_FILE` environment variable to the path of a YAML
```
dominique-roux: The variable is named LOG_CONFIG_FILE on fractal-web and fractal-data (both already merged).

tcompa: The corresponding variable here in the backend would rather be named `FRACTAL_..`, but that's not a big deal - I'll revert the change.

dominique-roux:
> cc @dominique-roux is there a reason for not including https://github.com/fractal-analytics-platform/fractal-server/blob/main/fractal_server/gunicorn_fractal.py in this PR?
> 
The log config file is not mandatory. If not configured FractalGunicornLogger is still used

tcompa:
> > cc @dominique-roux is there a reason for not including https://github.com/fractal-analytics-platform/fractal-server/blob/main/fractal_server/gunicorn_fractal.py in this PR?
> 
> The log config file is not mandatory. If not configured FractalGunicornLogger is still used

I'll rephrase, since I guess I was missing something: if the log-config-file variable is set, does this configuration also automatically apply to the [gunicorn logger](https://github.com/benoitc/gunicorn/blob/a8283bbf5e1416d0dd13f994f71ec2761988aeab/gunicorn/glogging.py#L164), or do [these defaults](https://github.com/benoitc/gunicorn/blob/a8283bbf5e1416d0dd13f994f71ec2761988aeab/gunicorn/glogging.py#L47-L85) apply?

(this is, of course, only relevant if you are using gunicorn)

dominique-roux:
> > cc @dominique-roux I committed [dfad2cd](https://github.com/fractal-analytics-platform/fractal-server/commit/dfad2cdc7cbd1a25d7d0230cad9cfe133467ecce) and then reverted it - so that we can discuss it.
> > Would this one be a viable solution?
> > The advantage is that we remove one of the "global" variables (which is not a great pattern, in my opinion), and we also remove the block handling `_CONFIG_ERROR` from `set_logger` (which is not the right place for logging information about startup errors).
> > Unrelated: it's not clear to me how the current implementation works. The comment says that the log is emitted only once, but what prevents emitting it again every time `set_logger` is called?
> 
> If you agree with this change, I'll then update the corresponding tests.

Yes agreed
I probably over-engineered a bit that because I wanted to make sure that the error in logger initialization would always be logged.
But now that I read again, using a logger.error there is enough

dominique-roux:
> > > cc @dominique-roux is there a reason for not including https://github.com/fractal-analytics-platform/fractal-server/blob/main/fractal_server/gunicorn_fractal.py in this PR?
> > 
> > 
> > The log config file is not mandatory. If not configured FractalGunicornLogger is still used
> 
> I'll rephrase, since I guess I was missing something: if the log-config-file variable is set, does this configuration also automatically apply to the [gunicorn logger](https://github.com/benoitc/gunicorn/blob/a8283bbf5e1416d0dd13f994f71ec2761988aeab/gunicorn/glogging.py#L164), or do [these defaults](https://github.com/benoitc/gunicorn/blob/a8283bbf5e1416d0dd13f994f71ec2761988aeab/gunicorn/glogging.py#L47-L85) apply?
> 
> (this is, of course, only relevant if you are using gunicorn)

if gunicorn logger is explicitly defined in in the yaml file it will use it otherwise it will use the code from the class

e.g.
```
version: 1
disable_existing_loggers: false

formatters:
  plain:
    format: "%(asctime)s [%(levelname)s] %(name)s (%(filename)s:%(lineno)d): %(message)s"
    datefmt: "%Y-%m-%d %H:%M:%S"

handlers:
  stdout:
    class: logging.StreamHandler
    stream: ext://sys.stdout
    formatter: plain
    level: INFO

  stderr:
    class: logging.StreamHandler
    stream: ext://sys.stderr
    formatter: plain
    level: ERROR

  debug_file:
    class: logging.handlers.RotatingFileHandler
    filename: /fractal-logs/fractal-server/debug.log
    maxBytes: 5242880   # 5 MB
    backupCount: 30
    formatter: plain
    level: DEBUG

  info_file:
    class: logging.handlers.RotatingFileHandler
    filename: /fractal-logs/fractal-server/info.log
    maxBytes: 5242880
    backupCount: 30
    formatter: plain
    level: INFO

  warning_file:
    class: logging.handlers.RotatingFileHandler
    filename: /fractal-logs/fractal-server/warning.log
    maxBytes: 5242880
    backupCount: 30
    formatter: plain
    level: WARNING

  error_file:
    class: logging.handlers.RotatingFileHandler
    filename: /fractal-logs/fractal-server/error.log
    maxBytes: 5242880
    backupCount: 30
    formatter: plain
    level: ERROR

  access_file:
    class: logging.handlers.RotatingFileHandler
    filename: /fractal-logs/fractal-server/access.log
    maxBytes: 5242880
    backupCount: 30
    formatter: plain
    level: INFO

loggers:
  fractal_server:
    level: DEBUG
    handlers: [stdout, stderr, debug_file, info_file, warning_file, error_file]
    propagate: false

  lifespan:
    level: DEBUG
    handlers: [stdout, stderr, debug_file, info_file, warning_file, error_file]
    propagate: false

  slow-response:
    level: DEBUG
    handlers: [stdout, stderr, debug_file, info_file, warning_file, error_file]
    propagate: false

  __main__:
    level: DEBUG
    handlers: [stdout, stderr, debug_file, info_file, warning_file, error_file]
    propagate: false

  uvicorn:
    level: INFO
    handlers: [stdout, stderr, info_file, error_file]
    propagate: false

  uvicorn.error:
    level: INFO
    handlers: [stdout, stderr, info_file, error_file]
    propagate: false

  uvicorn.access:
    level: INFO
    handlers: [stdout, access_file]
    propagate: false
  
  gunicorn:
    level: INFO
    handlers: [stdout, stderr, info_file, error_file]
    propagate: false

  gunicorn.error:
    level: INFO
    handlers: [stdout, stderr, info_file, error_file]
    propagate: false

  gunicorn.access:
    level: INFO
    handlers: [stdout, access_file]
    propagate: false

root:
  level: WARNING
  handlers: [stdout, stderr, warning_file, error_file]

```

tcompa:
> if gunicorn logger is explicitly defined in in the yaml file it will use it otherwise it will use the code from the class

Got it, thanks.

tcompa:
> > > cc @dominique-roux I committed [dfad2cd](https://github.com/fractal-analytics-platform/fractal-server/commit/dfad2cdc7cbd1a25d7d0230cad9cfe133467ecce) and then reverted it - so that we can discuss it.
> > > Would this one be a viable solution?
> > > The advantage is that we remove one of the "global" variables (which is not a great pattern, in my opinion), and we also remove the block handling `_CONFIG_ERROR` from `set_logger` (which is not the right place for logging information about startup errors).
> > > Unrelated: it's not clear to me how the current implementation works. The comment says that the log is emitted only once, but what prevents emitting it again every time `set_logger` is called?
> > 
> > 
> > If you agree with this change, I'll then update the corresponding tests.
> 
> Yes agreed I probably over-engineered a bit that because I wanted to make sure that the error in logger initialization would always be logged. But now that I read again, using a logger.error there is enough

Now in-place, see 13429a6a95d98ef7d9d729107874c92f885ab21b.

PR #3351, opened 2026-05-25
Highlight "WARNING"s from task logs

### LINKED ISSUE #3238: Highlight "WARNING"s from task logs
This is TBD (it's notably TBD whether it requires a backend-side feature or it can be implemented frontend-side)

ychiucco:
closes #3238 

- [x] I updated `run_task_compound` so that `has_warnings` is populated.
```bash
```


fractal_server/app/schemas/v2/job.py
```diff
@@ -116,6 +116,7 @@ class TaskStatusImages(BaseModel):
     num_submitted_images: int = 0
     num_done_images: int = 0
     num_failed_images: int = 0
+    num_warnings_images: int = 0
```
tcompa: We should name this `num_images_with_warnings`, even if it is very verbose.
This is both for consistency with units and to avoid thinking that `warnings` is another enum status like done/failed/submitted.

fractal_server/runner/executors/local/runner.py
```diff
@@ -29,6 +33,16 @@
...
 
 
+def _has_warnings(history_unit_id: int, db: Session) -> bool:
+    history_unit = db.get_one(HistoryUnit, history_unit_id)
+    grep = subprocess.run(  # nosec
+        ["grep", "-i", "WARNING", "-q", history_unit.logfile]
+    )
+    if grep.returncode == 0:
+        return True
+    return False
```
tcompa: Very minor style comment: I'd suggest to change this into one of the following:
1. Use `if/else` (with unit-test coverage)
2. `return (grep.returncode == 0)`
3. Add `check=True` and a try/except around the `run` call.

fractal_server/runner/executors/local/runner.py
```diff
@@ -179,6 +194,7 @@ def submit(
                     update_status_of_history_unit_no_commit(
                         history_unit_id=history_unit_id,
                         status=HistoryUnitStatus.DONE,
+                        has_warnings=_has_warnings(history_unit_id, db),
```
tcompa: Minor optimization comment:
Both `_has_warnings` and `update_status_of_history_unit_no_commit` make the same `get(HistoryUnit, history_unit_id)` db query. Should we change the interface a bit?

One option would be to add an `update_warnings: bool` argument to `update_status_of_history_unit_no_commit`, and then run the warnings-check from within that function.


**EDIT**: The same interface should be reviewed so that it also works for the SLURM case.

tcompa: Unrelated comment: this function should be renamed, based on its new functionality.

fractal_server/runner/executors/slurm_common/base_slurm_runner.py
```diff
@@ -518,16 +520,28 @@ def _postprocess_single_task(
         self,
         *,
         task: SlurmTask,
+        history_unit_id: int,
         was_job_scancelled: bool = False,
-    ) -> tuple[Any, Exception | None]:
+    ) -> tuple[Any, Exception | None, bool]:
```
tcompa: It'd be useful to have a docstring with the return-value description.

tcompa: Also: since `output` is the output of a `json.loads`, then we can be more specific about`result: Any` and rather use a JSON type like [this one](https://github.com/fractal-analytics-platform/fractal-task-tools/blob/b9a5d1b236be50d96dedb64e38e0faca074bbe82/src/fractal_task_tools/_json_types.py#L3-L5) - to be added into `fractal_server.types`.

fractal_server/app/routes/api/v2/history.py
```diff
@@ -114,17 +124,21 @@ async def get_history_run_list(
...
-            "num_submitted_units": 0,
-            "num_failed_units": 0,
-        }
-        for run_id in run_ids
-    }
     for run_id, unit_status, count in unit_counts:
         count_map[run_id][f"num_{unit_status}_units"] = count
 
+    stm = (
+        select(HistoryUnit.history_run_id, func.count(HistoryUnit.id))
```
tcompa, line 132: Filter out count=0 rows (not critical)

fractal_server/app/schemas/v2/job.py
```diff
@@ -117,6 +117,7 @@ class TaskStatusImages(BaseModel):
     num_done_images: int = 0
     num_failed_images: int = 0
     num_available_images: int | None = 0
+    num_images_with_warnings: int = 0
```
tcompa: Review whether all `0` defaults in this schema are actually needed.

tcompa: (add comment to explain)

fractal_server/runner/v2/db_tools.py
```diff
@@ -34,11 +35,17 @@ def update_status_of_history_unit_no_commit(
     history_unit_id: int,
     status: HistoryUnitStatus,
     db_sync: Session,
+    update_warnings: bool = False,
```
tcompa: Line below: introduce `get_one`

fractal_server/runner/v2/db_tools.py
```diff
@@ -34,11 +35,17 @@ def update_status_of_history_unit_no_commit(
...
 ) -> None:
     unit = db_sync.get(HistoryUnit, history_unit_id)
     if unit is None:
         raise ValueError(f"HistoryUnit {history_unit_id} not found.")
     unit.status = status
+    if update_warnings:
+        grep = subprocess.run(  # nosec
+            ["grep", "-i", "WARNING", "-q", unit.logfile]
```
tcompa: Add grep to https://fractal-analytics-platform.github.io/fractal-server/install_and_deploy/

fractal_server/runner/executors/local/runner.py
```diff
@@ -180,6 +181,7 @@ def submit(
                         history_unit_id=history_unit_id,
                         status=HistoryUnitStatus.DONE,
                         db_sync=db,
+                        update_warnings=True,
```
tcompa: rename function, without `status_of`

fractal_server/runner/executors/slurm_common/base_slurm_runner.py
```diff
@@ -810,6 +833,7 @@ def submit(
                                     history_unit_id=history_unit_id,
                                     status=HistoryUnitStatus.DONE,
                                     db_sync=db,
+                                    update_warnings=True,
```
tcompa: line 851: update warnings

fractal_server/runner/v2/db_tools.py
tcompa: Remove argument and branch, make it always happen

fractal_server/runner/v2/runner_functions.py
```diff
@@ -602,15 +605,37 @@ def run_task_compound(
...
+                ["grep", "-i", "WARNING", "-q", logfile]
+            ).returncode
+            == 0
+        ]
+        db.execute(
+            update(HistoryUnit)
+            .where(HistoryUnit.id.in_(units_with_warnings))
+            .values(has_warnings=True)
+        )
+        db.commit()
```
tcompa: add batching - like in the upsert function

fractal_server/app/routes/api/v2/history.py
```diff
@@ -280,14 +295,25 @@ async def get_history_images(
         attribute_filters=request_body.attribute_filters,
     )
 
+    if include_warnings:
```
tcompa, line 298: Add test

fractal_server/app/routes/api/v2/latest_job.py
```diff
@@ -156,6 +156,20 @@ async def get_latest_job(
...
+                HistoryImageCache.latest_history_unit_id == HistoryUnit.id,
+            )
+            .where(HistoryImageCache.dataset_id == dataset_id)
+            .where(HistoryImageCache.workflowtask_id == wftask.id)
+            .where(HistoryUnit.has_warnings.is_(True))
+        )
+        res = await db.execute(stm)
+        warnings_count = res.scalar()
+        setattr(statuses[wftask.id], "num_images_with_warnings", warnings_count)
+
```
tcompa: 1. `num_images_with_warnings` should become a boolean flag (`has_warnings` or similar naming).
2. The query should not count all images/units with warnings, but only find out whether at least one exists.
3. Same change needs to happen in the frontend.
4. Base on task type, perform different queries:

For compound/converter-compound tasks: use a different query, that ignores `HistoryImageCache` and  directly queries `HistoryUnit`s and `HistoryRun`s - such that the run FKs match with the current dataset/wftask IDs and the units have/don't have warnings.

For the others: keep the current query.

tcompa: something like
```
select(unit)
where(unit.has_warnings)
join(run)
where(run.wftask_id=..)
where(run.dataset_id=..)
```

```

tcompa:
Running `test_bulk_update_has_warnings_history_unit` with `23_456` units takes ~18 seconds, locally.
Adding internal profiling to the `bulk_update_has_warnings_history_unit` function, it turns out that the vast majority of that time is spent in 23k `subprocess.run` calls (for a total of ~17.6 seconds).

tcompa:
> Running `test_bulk_update_has_warnings_history_unit` with `23_456` units takes ~18 seconds, locally. Adding internal profiling to the `bulk_update_has_warnings_history_unit` function, it turns out that the vast majority of that time is spent in 23k `subprocess.run` calls (for a total of ~17.6 seconds).

Precomputing the grep executable (with `shutil.which("grep")`) seems to reduce the 16-18 seconds to 13-14 seconds.

fractal_server/app/routes/api/v2/history.py
tcompa: Already taken care of by the original query.

Issue #2285, opened 2025-02-18
Acquire history-item row lock before making granular updates to image statuses


tcompa:
Ref:
* https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.CompoundSelect.with_for_update
* https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS

PR #3360, opened 2026-06-09
Introduce core tasks

### LINKED ISSUE #3163: Allow admins to tag tasks
A discussion that has come up a few times and I just had with @schulz-m again is about how we mark certain task as "core", i.e. "you can expect to get support for this", vs. community tasks that the Fractal admins on a given cluster can't make guarantees for.

We've had this discussion in many directions before, even towards having validation standards for tasks. In the brainstorming with Max, we came to a more narrowly scoped definition that may be quite helpful:

Some user stories:
As an admin, I can mark/tag specific tasks (/task packages) as "core". This shows up in some visual ways for users. In that way, on every server, the admins can signal which task packages are best supported, where users can expect support etc.

As a user, I can see a visual indicator on the "Add new workflow task" page & the task list page about which tasks are core, e.g. by showing a star symbol next to them.

As a user, I'm guided to use core tasks first. For example, they show up first in the task list. (easier if it's for whole task packages rather than individual tasks)

As a user, I can filter for certain task "tags".

### Open discussion points:

- Does one star specific tasks or whole task packages? Would be simpler to manage it on the task package level, but if the per task nuance is needed, it would be buildable
- Are those tags per task version or do they apply to all versions? It would be potentially suboptimal that I loose the core tag when installing an updated version of a task package & always having to retag them
- Is a single tag sufficient or do there need to be categories of tasks, e.g. "core", "validated", "experimental", "community"
- Can a task have a single tag or multiple tags?

This is different from tags in the manifest, which are statically defined by task developers. These marks/tags here are meant for an admin of a server to signal to the given users on that instance

ychiucco:
closes #3163 

```bash
```


fractal_server/app/routes/admin/v2/task.py
```diff
@@ -161,3 +166,65 @@ async def query_tasks(
...
+        .where(TaskGroupV2.version == task_group.version)
+        .where(TaskGroupV2.resource_id == task_group.resource_id)
+        .where(TaskV2.name == task.name)
+        .where(TaskV2.is_core.is_(True))
+    )
+    if res.scalars().all():
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail="TBD",
+        )
```
tcompa: This should later end up in an auxiliary function similar to `_verify_non_duplication_group_constraint`.

fractal_server/app/routes/api/v2/_aux_functions_tasks.py
```diff
@@ -391,3 +391,26 @@ def _check_type_filters_compatibility(
...
+        select(TaskV2.id)
+        .join(TaskGroupV2, TaskGroupV2.id == TaskV2.taskgroupv2_id)
+        .where(TaskGroupV2.pkg_name == task_group.pkg_name)
+        .where(TaskGroupV2.version == task_group.version)
+        .where(TaskGroupV2.resource_id == task_group.resource_id)
+        .where(TaskV2.name == task.name)
+        .where(TaskV2.is_core.is_(True))
+    )
+    duplicate_task_id = res.scalar_one_or_none()
+    if duplicate_task_id and duplicate_task_id != task.id:
```
tcompa: > `duplicate_task_id != task.id`

Could we include this check in the sql query?

fractal_server/app/routes/api/v2/_aux_functions_tasks.py
```diff
@@ -391,3 +391,27 @@ def _check_type_filters_compatibility(
...
+    res = await db.execute(
+        select(TaskV2.id)
+        .where(TaskV2.name == task.name)
+        .where(TaskV2.id != task.id)
+        .where(TaskV2.is_core.is_(True))
+        .join(TaskGroupV2, TaskGroupV2.id == TaskV2.taskgroupv2_id)
+        .where(TaskGroupV2.pkg_name == task_group.pkg_name)
+        .where(TaskGroupV2.version == task_group.version)
+        .where(TaskGroupV2.resource_id == task_group.resource_id)
+    )
```
tcompa: If we add `limit(1)`, we make sure that the query will stop at the first hit (if any).

fractal_server/app/routes/api/v2/_aux_functions_tasks.py
```diff
@@ -391,3 +391,27 @@ def _check_type_filters_compatibility(
...
+    duplicate_task_id = res.scalar_one_or_none()
+    if duplicate_task_id:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail="TBD",
```
tcompa: This is an admin-facing message, which could read something like
```python
"There already exists a core task with pkg_name={task_group.pkg_name}, version={task_group.version} and name='{task.name}' (task ID: {duplicate_task_id})."
```

tests/v2/test_03_api/admin/test_admin_task.py
```diff
@@ -0,0 +1,277 @@
...
+    tg = TaskGroupV2(
+        user_id=user_id,
+        resource_id=resource.id,
+        origin="unknown",
+        pkg_name="foo",
+        version="1",
+        task_list=[task_a, task_a_copy, task_b],
+    )
+    db.add(tg)
+    await db.commit()
```
tcompa: This status (two tasks with the same name in the same task group) should not be reachable through the API. TBD whether it is currently possible to reach it.

tcompa: See https://github.com/fractal-analytics-platform/fractal-server/issues/3361

fractal_server/app/routes/admin/v2/task.py
```diff
@@ -161,3 +174,66 @@ async def query_tasks(
...
         items=task_info_list,
     )
+
+
+@router.post(
+    "/make-core/",
+    status_code=status.HTTP_200_OK,
+)
+async def make_task_core(
+    task_ids: list[int],
```
tcompa: This list should be made of unique elements

fractal_server/app/routes/admin/v2/task.py
```diff
@@ -161,3 +174,66 @@ async def query_tasks(
...
+
+    return Response(status_code=status.HTTP_200_OK)
+
+
+@router.post(
+    "/make-not-core/",
+    status_code=status.HTTP_200_OK,
+)
+async def make_task_not_core(
+    task_ids: list[int],
```
tcompa: This list should be made of unique elements

fractal_server/app/routes/admin/v2/task.py
```diff
@@ -161,3 +174,66 @@ async def query_tasks(
...
+            await _verify_non_duplication_task_core_constraint(
+                task=task, task_group=task_group, db=db
+            )
+            task.is_core = True
+    except Exception as e:
+        for task, is_core in zip(tasks, current_status):
+            task.is_core = is_core
+        db.add_all(tasks)
+        await db.commit()
+        raise e
```
tcompa: It'd be nice to avoid the "commit and then revert with another commit" pattern, let's explore some options.

The one that comes to mind is the following:
1. For all task+taskgroup pairs from the request body, run `_verify_non_duplication_task_core_constraint` (as already done in the block above).
2. Run a validity check on the task+taskgroup pairs from the request body, by checking that their `(task.name, task_group.pkg_name, task_group.version, task_group.resource_id)` list only has unique elements.
3. If both checks 1 and 2 passed, proceed with a single `update` db operation.


This is clearly at risk for race conditions, but the current version also is. I think we can fix it with https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.Select.with_for_update - see https://github.com/fractal-analytics-platform/fractal-server/issues/2285#issuecomment-2667822511.

fractal_server/app/routes/api/v2/task.py
```diff
@@ -59,6 +60,7 @@ async def get_list_task(
     category: str | None = None,
     modality: str | None = None,
     author: str | None = None,
+    only_core: bool = False,
```
tcompa: Remove this

fractal_server/app/routes/admin/v2/task.py
```diff
@@ -161,3 +176,67 @@ async def query_tasks(
...
+    ]
+    if len(set(payload_tuples)) != len(payload_tuples):
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail="TBD: duplication in payload",
+        )
+    for task in tasks:
+        task.is_core = True
+    db.add_all(tasks)
+    await db.commit()
```
tcompa: What about `update(TaskV2, values=..).where(TaskV2.id.in_(...))`?
The current form was meant to keep consistency between the two endpoints, but now both endpoints can switch to `update`.

fractal_server/app/routes/admin/v2/task.py
```diff
@@ -161,3 +176,67 @@ async def query_tasks(
...
+    task_ids: ListUniqueNonNegativeInt,
+    superuser: UserOAuth = Depends(current_superuser_act),
+    db: AsyncSession = Depends(get_async_db),
+) -> Response:
+    tasks = [
+        await _get_task_or_404(task_id=task_id, db=db) for task_id in task_ids
+    ]
+    task_groups = [
+        await db.get_one(TaskGroupV2, task.taskgroupv2_id) for task in tasks
+    ]
```
tcompa:
```python
res = execute(select(TaskV2,TaskGroupV2).join(...).where(TaskV2.id.in_(task_ids))
tasks = res.scalars()
len(tasks) != len(task_ids):
    404 (more or less precise - TBD)
```

fractal_server/app/routes/admin/v2/task.py
```diff
@@ -161,3 +176,67 @@ async def query_tasks(
...
+            task_group.pkg_name,
+            task_group.version,
+            task_group.resource_id,
+        )
+        for task, task_group in zip(tasks, task_groups)
+    ]
+    if len(set(payload_tuples)) != len(payload_tuples):
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail="TBD: duplication in payload",
```
tcompa: > This request would generate conflicting core tasks (with the same task name and task-group properties).
> Hint: include fewer tasks in the request body and retry.

fractal_server/app/routes/admin/v2/task.py
tcompa: "with lock for update"?

fractal_server/app/routes/admin/v2/task.py
```diff
@@ -161,3 +176,67 @@ async def query_tasks(
...
+    db: AsyncSession = Depends(get_async_db),
+) -> Response:
+    tasks = [
+        await _get_task_or_404(task_id=task_id, db=db) for task_id in task_ids
+    ]
+
+    for task in tasks:
+        task.is_core = False
+    db.add_all(tasks)
+    await db.commit()
```
tcompa: As in the previous endpoint:
* get through sql
* update through sql

fractal_server/app/routes/admin/v2/task.py
```diff
@@ -16,12 +20,16 @@
 from fractal_server.app.models.v2 import TaskV2
 from fractal_server.app.models.v2 import WorkflowTaskV2
 from fractal_server.app.models.v2 import WorkflowV2
+from fractal_server.app.routes.api.v2._aux_functions_tasks import (
```
tcompa: This function should be moved to a module within `app/routes/admin`, since it's not used from `app/routes/api/v2`

PR #3385, opened 2026-06-30
Remove `slim` from `GET /api/v2/task-group/`

### LINKED ISSUE #3384: Review whether any client uses `slim=False` for `GET /api/v2/task-group/`

ychiucco:
closes #3384 

```bash
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 # Unreleased
 
+* API:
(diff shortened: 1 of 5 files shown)


tcompa:
cc @dominique-roux this PR removes a query parameter, and the default behavior changes from including many task attributes to including only a few (notably dropping e.g. the JSON schemas, to reduce the response-body size).

Can you please double check if this would break any existing use case?

tcompa (approved):

PR #3387, opened 2026-07-01
Support make entire task-group core

### LINKED ISSUE #3386: Support "mark an entire task group as core"
The first plan would be to go with:
- `POST /admin/v2/task-group/{task_group_id}/make-core/`,
- `POST /admin/v2/task-group/{task_group_id}/make-not-core/`.

ychiucco:
closes #3386 

```bash
```


fractal_server/app/routes/admin/v2/_aux_functions_core_tasks.py
```diff
@@ -34,3 +36,93 @@ async def _verify_non_duplication_task_core_constraint(
                 f"and name='{task.name}' (task ID: {duplicate_task_id})."
             ),
         )
+
+
+async def _make_task_core_bulk(
+    *,
+    task_ids: ListUniqueNonNegativeInt,
+    db: AsyncSession,
+) -> None:
```
tcompa: Docstring

fractal_server/app/routes/admin/v2/_aux_functions_core_tasks.py
```diff
@@ -34,3 +36,93 @@ async def _verify_non_duplication_task_core_constraint(
...
+    res = await db.execute(
+        select(TaskV2, TaskGroupV2)
+        .join(TaskGroupV2, TaskGroupV2.id == TaskV2.taskgroupv2_id)
+        .where(TaskV2.id.in_(task_ids))
+    )
+    tasks_and_groups = res.all()
```
tcompa: Note that this is slightly suboptimal when we call this function for all the tasks of a given (known) task group.
It's OK for the current endpoint (which only concerns a single task group), but it would not be reasonable if we introduced a bulk make-many-task-groups-core endpoint.

fractal_server/app/routes/admin/v2/_aux_functions_core_tasks.py
```diff
@@ -34,3 +36,93 @@ async def _verify_non_duplication_task_core_constraint(
...
+    if len(tasks_and_groups) != len(task_ids):
+        missing_ids = sorted(
+            list(set(task_ids) - set([tg[0].id for tg in tasks_and_groups]))
+        )
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Not all tasks were found (Missing IDs: {missing_ids}).",
+        )
```
tcompa: Note that this is slightly suboptimal when we call this function for all the tasks of a given (known) task group.
It's OK for the current endpoint (which only concerns a single task group), but it would not be reasonable if we introduced a bulk make-many-task-groups-core endpoint.

fractal_server/app/routes/admin/v2/_aux_functions_core_tasks.py
```diff
@@ -34,3 +36,93 @@ async def _verify_non_duplication_task_core_constraint(
...
+    ]
+    if len(set(payload_tuples)) != len(payload_tuples):
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail=(
+                "This request would generate conflicting core tasks "
+                "(with the same task name and task-group properties). "
+                "Hint: include fewer tasks in the request body and retry."
+            ),
+        )
```
tcompa: Note that also this check is redundant, when calling the task-group endpoint.

fractal_server/app/routes/admin/v2/_aux_functions_core_tasks.py
```diff
@@ -34,3 +36,93 @@ async def _verify_non_duplication_task_core_constraint(
...
+            detail=f"Not all tasks were found (Missing IDs: {missing_ids}).",
+        )
+
+    # Update
+    await db.execute(
+        update(TaskV2).where(TaskV2.id.in_(task_ids)).values(is_core=False)
+    )
+    await db.commit()
+
+    return
```
tcompa: I suggest we remove this function, since its usage in the task-group endpoint could be replaced with
```python
await db.execute(update(TaskV2).where(TaskV2.taskgroupv2_id == something).values(is_core=False))
await db.commit()
```
which would be more readable (on top of being more efficient).

tcompa: (up to a preliminary task-group existence check, with a 404 branch)

tests/v2/test_03_api/admin/test_admin_task.py
```diff
@@ -313,6 +313,12 @@ async def test_task_core(
     assert task_b.is_core is False
 
     async with MockCurrentUser(is_superuser=True):
+        # Test empty list
+        res = await client.post(
+            f"{PREFIX}/task/make-core/",
+            json=[],
+        )
+        assert res.status_code == 200
```
tcompa: I suggest we make this response a 422, so that if the client has a bug and sends an empty list at least we catch it earlier.

tests/v2/test_03_api/admin/test_admin_taskgroup.py
```diff
@@ -742,3 +743,54 @@ async def test_admin_delete_task_group_api_ssh(
...
+        )
+        assert res.status_code == 200
+
+        await db.refresh(task1)
+        await db.refresh(task2)
+        assert task1.is_core is False
+        assert task2.is_core is False
+
+        res = await client.post(f"{PREFIX}/task-group/123456789/make-not-core/")
+        assert res.status_code == 200
```
tcompa: I would rather expect a 404 here, the same as for `POST /api/v2/task/make-core/` with payload `[1232134234234234, 2342342342342]`.

fractal_server/app/routes/admin/v2/task.py
```diff
@@ -292,16 +292,18 @@ async def make_task_core(
         update(TaskV2).where(TaskV2.id.in_(task_ids)).values(is_core=True)
     )
     await db.commit()
-
-    return Response(status_code=status.HTTP_200_OK)
+    return Response(
+        content=f"{len(task_ids)} tasks have been made core.",
```
tcompa: Is this message displayed in the web-client? If not, let's drop it.

tcompa (approved):

PR #3401, opened 2026-07-10
Include `is_core` in `AvailableTask`

### LINKED ISSUE #3400: Include `is_core` in `AvailableTask` for workflow import
To be added in 

https://github.com/fractal-analytics-platform/fractal-server/blob/206a20bd46e67a132593d5e48dfc8d8aafd222d8/fractal_server/app/routes/api/v2/workflow_import.py#L53-L61

and then computed/propagated in the workflow-import endpoint

ychiucco:
closes #3400

```bash
```


fractal_server/app/routes/api/v2/workflow_import.py
```diff
@@ -174,6 +175,13 @@ async def _get_task_id_or_available_tasks(
                         < _version_sort_key(task_import.version)
                     ),
                     active=tg.active,
+                    is_core=(
+                        next(
+                            t.is_core
+                            for t in tg.task_list
+                            if t.name == task_import.name
+                        )
+                    ),
```
tcompa: I was puzzled at first, but then I think I understood this: the guarantee that `next` will not raise a `StopIteration` comes from the filtering taking place at lines 133-140. Is this correct? If so, let's add a comment about it because if one only looked at this code part it is quite hard to understand it - in my opinion.

PR #3404, opened 2026-07-13
Sync-core-tasks feature

### LINKED ISSUE #3402: Synchronize core-tasks list
This is a feature to be added as a CLI command (e.g. `fractalctl sync-core-tasks`).

### (1) Public core-tasks list

The first (required) input is a public list of tasks to be considered core.

This will be available at a given URL (e.g. https://raw.githubusercontent.com/fractal-analytics-platform/fractal-analytics-platform.github.io/refs/heads/main/core-tasks-list.json). The precise structure can be decided based on how it is used for the current feature.

<details>

<summary> Expand to see examples </summary>

Example 1:
```json
{
    "fractal-tasks-core": {
        "2.0.1": [
            "Illumination Correction",
            ],
        "2.0.2": [
            "Illumination Correction",
            "Threshold Segmentation",
            ]
    },
    "fractal-cellpose-sam-task": {
        "0.2.2": [
            "Cellpose SAM Segmentation",
        ]
    }
}
```

Example 2:
```json
[
    {
        "pkg_name": "fractal-tasks-core",
        "version": "2.0.1",
        "task_name": "Illumination Correction",
    },
    {
        "pkg_name": "fractal-tasks-core",
        "version": "2.0.2",
        "task_name": "Illumination Correction",
    },
    {
        "pkg_name": "fractal-tasks-core",
        "version": "2.0.2",
        "task_name": "Threshold Segmentation",
    },
    {
        "pkg_name": "fractal-cellpose-sam-task",
        "version": "0.2.2",
        "task_name": "Cellpose SAM Segmentation",
    },
]
```

Example 3:
```json
[
    ["fractal-tasks-core", "2.0.1", "Illumination Correction"],
    ["fractal-tasks-core", "2.0.2", "Illumination Correction"],
    ["fractal-tasks-core", "2.0.2", "Threshold Segmentation"],
    ["fractal-cellpose-sam-task", "0.2.2", "Cellpose SAM Segmentation"],
]
```

</details>

Also TBD is whether the CLI command reads the URL directly, or reads it from an on-disk file (which was updated via `curl` or `wget` beforehand). Reading from URL is simpler, but reading from disk would be more homogeneous with the other inputs (see below).

Let's start with JSON files for all inputs, and later we'll move the public list to URL-fetching and the customizations to a new database table. The paths can be passed as CLI arguments.

### (2) Resource/group customization

A instance-specific JSON file like this
```json
[
    {"resource_id": 1, "user_group_id": 6},
    {"resource_id": 2, "user_group_id": 7},
    {"resource_id": 3, "user_group_id": 3},
]
```
which says "On resource 1, only tasks shared with user group 6 should be marked as core", ...

We would start with having this on-disk, and passing it to the CLI command.

### (3) Additional customizations

Customizations to input 1 (the public core-tasks list) can be of two types.

Say that the public list looks like
```json
[
    ["fractal-tasks-core", "2.0.1", "Illumination Correction"],
    ["fractal-tasks-core", "2.0.2", "Illumination Correction"],
    ["fractal-tasks-core", "2.0.2", "Threshold Segmentation"],
    ["fractal-cellpose-sam-task", "0.2.2", "Cellpose SAM Segmentation"],
]
```
then we could have `ignore.json` looking like
```json
[
    ["fractal-tasks-core", "2.0.2", "Illumination Correction"],
    ["fractal-tasks-core", "2.0.2", "Threshold Segmentation"],
]
```
e.g. because `fractal-tasks-core` v2.0.2 cannot be installed on a given instance.

And we could also have `additions.json`, like
```json
[
    ["another-task-group", "1.2.3", "Task Name"]
]
```
to be added only on the specific instance.

TBD whether these customizations are stored on-disk or they are stored in the database (-> we'll start on-disk, and likely move to the db later).

If we implement these two customizations, we can make a full synchronization (meaning that a task which was not listed either in the public list or in the per-instance customization list won't be core, even if an admin marked it as core in the past).

## CLI signature

It could look like
```bash
fractalctl sync-core-tasks --main-list /tmp/core.json --ignore-list /tmp/ignore.json --add-list /tmp/additions.json --groups /tmp/resource-usergroup.json
```

## Implementation

1. Validate that the resource and usergroup IDs point to esisting db objects.
2. Create a single core-tasks list, by integrating the different inputs.
3. Validate the core-tasks list, and fail if some item is not valid or if it includes some duplications.
4. Start a nested db transaction, so that we can rollback upon any exception.
5. Mark all tasks as non-core (without committing).
6. Loop over the core-tasks list, and flag them as core.
7. If a task is not found, continue. This is not a blocking issue.
8. Commit, as a single transaction
9. If any error takes place (during the loop or upon committing), roll back to the initial state and log the failure (including a non-zero exit code).
10. If the instance has emails configured, send a warning email with the error logs.

## Automation

The script to run in cron would look like
```
#!/bin/bash

set -eu

curl https://example.org/list.json /backend-dir/core/core.json  # with options for overwriting

fractalctl sync-core-tasks \
    --main-list /backend-dir/core/core.json \
    --ignore-list /backend-dir/core/ignore.json \
    --add-list /backend-dir/core/additions.json \
    --groups /backend-dir/core/resource-usergroup.json
```

tcompa:

```bash
```


fractal_server/cli/_sync_core_tasks.py
```diff
@@ -0,0 +1,136 @@
...
+from fractal_server.app.models import Resource
+from fractal_server.app.models import TaskGroupV2
+from fractal_server.app.models import TaskV2
+from fractal_server.app.models import UserGroup
+from fractal_server.config import get_email_settings
+from fractal_server.send_mail import send_fractal_email_or_log_failure
+from fractal_server.syringe import Inject
+from fractal_server.types import NonEmptyStr
+
+_ThreeStringsTuple = tuple[
```
ychiucco: Why not directly `SetThreeStringsTuples = set[tuple[NonEmptyStr, NonEmptyStr, NonEmptyStr]`?

fractal_server/cli/_sync_core_tasks.py
```diff
@@ -0,0 +1,136 @@
...
+
+def sync_core_tasks(
+    *,
+    resources_and_groups: Path,
+    base: Path | None = None,
+    add: Path | None = None,
+    remove: Path | None = None,
+):
+    """
+    FIXME
```
ychiucco: fix it

ychiucco (approved):

fractal_server/cli/_sync_core_tasks.py
tcompa: Done

fractal_server/cli/_sync_core_tasks.py
tcompa: thanks - fixed

PR #3416, opened 2026-07-21
Add /job/squeue/ endpoint

### LINKED ISSUE #3407: Add `squeue` endpoint
The main goal is to expose a new endpoint that calls `squeue` (https://slurm.schedmd.com/squeue.html) on the cluster and returns its stdout (or a 422 response if the command fails). Note that this is _not_ the first step towards a full integration of SLURM info into Fractal (which would require a much broader discussion and planning), but rather a single standalone feature

## Endpoint

The endpoint would be `GET /api/v2/job/squeue/`, with a single query parameter taking values in `["all", "user", "accounts"]`.

As a first check, the endpoint only works for the two SLURM backends, while it should respond with some 400 or 422 error for a local backend.
The second check is the [validation of the user profile](https://github.com/fractal-analytics-platform/fractal-server/blob/7aedb393d3944fdbdc02ca9d8cac4900b546fe9e/fractal_server/app/routes/api/v2/submit.py#L145-L149).

Then the value of the query parameter determines the precise command to run:
```python
# all
cmd = "squeue"

# user
cmd = "squeue --user user1"

# accounts
cmd = "squeue --accounts=account1,account2,account3"
```
where
* In the `user` case, the user is the one defined in `Profile.username`
* In the `accounts` case, the list of accounts is the one in `UserOAuth.slurm_accounts` (for the user calling the endpoint). If the list is empty, we fall back to `squeue`.

> Note: we should also add a `--format` option, but I still need to play with it a bit to find a reasonable one.

The command call should happen differently based on the backend:
* For a SSH backend, it should create a new single-use connection (as done [here](https://github.com/fractal-analytics-platform/fractal-server/blob/7aedb393d3944fdbdc02ca9d8cac4900b546fe9e/fractal_server/tasks/v2/ssh/collect.py#L88-L95)), and then call the [`run_command` method](https://github.com/fractal-analytics-platform/fractal-server/blob/7aedb393d3944fdbdc02ca9d8cac4900b546fe9e/fractal_server/ssh/_fabric.py#L361).
* On a sudo-slurm backend, this should go through `subprocess.run`. We have a wrapper available at https://github.com/fractal-analytics-platform/fractal-server/blob/7aedb393d3944fdbdc02ca9d8cac4900b546fe9e/fractal_server/utils.py, but I think it would actually make things more complex than a direct `subprocess.run` call.

If the command fails, we should catch and log the exception and then return a placeholder text (not the actual exception).
If the command goes through, we should return a 200 response with the stdout.

## Tests

The test for the SSH case could be similar to https://github.com/fractal-analytics-platform/fractal-server/blob/7aedb393d3944fdbdc02ca9d8cac4900b546fe9e/tests/v2/test_08_backends/test_slurm_ssh_squeue.py#L14, with the calls to the runner function `run_squeue` replaced by calls to the API. Note that we cannot use the runner function from the API, as the runner is an object which only exists in a background job (and also because the squeue format would be slightly different).

The test for the sudo-SLURM case should include the `monkey_slurm` and `slurm_sudo_resource_profile_db` fixtures. The structure could be similar to [this one](https://github.com/fractal-analytics-platform/fractal-server/blob/7aedb393d3944fdbdc02ca9d8cac4900b546fe9e/tests/v2/test_08_backends/test_slurm_sudo_runner.py#L30), but it's also perfectly fine if we only test if for a dummy `squeue` output (if there are no jobs, `squeue` will only return the header line). This actually also holds in general (including the SSH case): we need to test that the endpoint runs the correct command, but we don't really care about testing the `squeue` output.

zonia3000:

```bash
```

Closes #3407

### DIFF
--- a/fractal_server/app/routes/api/v2/job.py
+++ b/fractal_server/app/routes/api/v2/job.py
@@ -277,7 +277,7 @@ async def get_squeue(
 
     command = (
         f"squeue {flags} --format="
-        f'"%.18i %.9P %.20j %.14u %.14a %.11T %.12M %.6D %R"'
+        f'"%.12i %.9P %.24j %.14u %.14a %.11T %.12M %.6D %.4C %.10m %R"'
     )
 
     try:
@@ -299,5 +299,5 @@ async def get_squeue(
         logger.error(f"Cannot execute squeue command. Original error: {str(e)}")
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
-            detail="Error executing squeue command.",
+            detail="Error executing squeue command - please retry later.",
         )
--- a/fractal_server/app/routes/api/v2/job.py
+++ b/fractal_server/app/routes/api/v2/job.py
@@ -34,9 +34,8 @@ from fractal_server.app.schemas.v2 import JobStatusType
 from fractal_server.app.schemas.v2 import ResourceType
 from fractal_server.app.schemas.v2.sharing import ProjectPermissions
 from fractal_server.config import get_settings
-from fractal_server.logger import get_logger
+from fractal_server.logger import set_logger
 from fractal_server.runner.filenames import WORKFLOW_LOG_FILENAME
-from fractal_server.ssh._fabric import FractalSSHCommandError
 from fractal_server.ssh._fabric import SingleUseFractalSSH
 from fractal_server.ssh._fabric import SSHConfig
 from fractal_server.syringe import Inject
@@ -56,6 +55,8 @@ async def zip_folder_threaded(folder: str) -> Iterator[bytes]:
 
 
 router = APIRouter()
+
+logger = set_logger(__name__)
 
 
 @router.get("/job/", response_model=list[JobRead])
@@ -254,8 +255,6 @@ async def get_squeue(
     user: UserOAuth = Depends(get_api_user),
     db: AsyncSession = Depends(get_async_db),
 ):
-    logger = get_logger()
-
     settings = Inject(get_settings)
     backend = settings.FRACTAL_RUNNER_BACKEND
     if backend not in [ResourceType.SLURM_SUDO, ResourceType.SLURM_SSH]:
@@ -281,46 +280,24 @@ async def get_squeue(
         f'"%.18i %.9P %.20j %.14u %.14a %.11T %.12M %.6D %R"'
     )
 
-    if backend == ResourceType.SLURM_SSH:
-        with SingleUseFractalSSH(
-            ssh_config=SSHConfig(
-                host=resource.host,
-                user=profile.username,
-                key_path=profile.ssh_key_path,
-            ),
-            logger_name=logger.name,
-        ) as fractal_ssh:
-            ssh_ok = True
-            try:
-                fractal_ssh.check_connection()
-            except Exception as e:
-                ssh_ok = False
-                logger.error(
-                    f"Cannot establish SSH connection. Original error: {str(e)}"
-                )
-
-            if not ssh_ok:
-                raise HTTPException(
-                    status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
-                    detail="Cannot establish SSH connection.",
-                )
-
-            try:
+    try:
+        if resource.type == ResourceType.SLURM_SSH:
+            with SingleUseFractalSSH(
+                ssh_config=SSHConfig(
+                    host=resource.host,
+                    user=profile.username,
+                    key_path=profile.ssh_key_path,
+                ),
+                logger_name=logger.name,
+            ) as fractal_ssh:
                 out = fractal_ssh.run_command(cmd=command)
                 return PlainTextResponse(content=out)
-            except FractalSSHCommandError as e:
-                logger.error(f"squeue command failed. Original error: {str(e)}")
-                raise HTTPException(
-                    status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
-                    detail="squeue command failed.",
-                )
-    else:
-        try:
+        else:
             out = execute_command_sync(command=command)
             return PlainTextResponse(content=out)
-        except Exception as e:
-            logger.error(f"squeue command failed. Original error: {str(e)}")
-            raise HTTPException(
-                status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
-                detail="squeue command failed.",
-            )
+    except Exception as e:
+        logger.error(f"Cannot execute squeue command. Original error: {str(e)}")
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail="Error executing squeue command.",
+        )
--- a/fractal_server/app/routes/api/v2/job.py
+++ b/fractal_server/app/routes/api/v2/job.py
@@ -1,12 +1,16 @@
 import asyncio
 from collections.abc import Iterator
 from pathlib import Path
+from typing import Literal
 from typing import Sequence
 
 from fastapi import APIRouter
 from fastapi import Depends
+from fastapi import HTTPException
 from fastapi import Response
 from fastapi import status
+from fastapi.params import Query
+from fastapi.responses import PlainTextResponse
 from fastapi.responses import StreamingResponse
 from sqlmodel import select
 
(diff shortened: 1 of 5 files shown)


fractal_server/app/routes/api/v2/job.py
```diff
@@ -228,3 +243,84 @@ async def stop_job(
...
+    resource, profile = await validate_user_profile(
+        user=user,
+        db=db,
+    )
+
+    flags = ""
+    if scope == "user":
+        flags = f"--user {profile.username}"
+    elif scope == "accounts" and len(user.slurm_accounts) > 0:
+        flags = f"--accounts={','.join(user.slurm_accounts)}"
```
zonia3000, line 276: The `slurm_accounts` values can be set by the user from the "My profile" page. I noticed that there is no special validation on these values, so a malicious user could set them to a value that could manipulate this command. Is this something that we should consider?

A simple example that will cause issues here could be:

```
PATCH /api/auth/current-user

{"slurm_accounts":["test01","x\n ls -l /\necho"]}
```

In this case the modal will show both the `squeu`e output and the output of `ls -l /`.

I think that we can restrict which characters we consider valid for slurm account names.

fractal_server/app/routes/api/v2/job.py
```diff
@@ -228,3 +243,84 @@ async def stop_job(
...
+    )
+
+    if backend == ResourceType.SLURM_SSH:
+        with SingleUseFractalSSH(
+            ssh_config=SSHConfig(
+                host=resource.host,
+                user=profile.username,
+                key_path=profile.ssh_key_path,
+            ),
+            logger_name=logger.name,
```
zonia3000: Is this name ok?

fractal_server/app/routes/api/v2/job.py
```diff
@@ -228,3 +243,84 @@ async def stop_job(
...
+        flags = f"--user {profile.username}"
+    elif scope == "accounts" and len(user.slurm_accounts) > 0:
+        flags = f"--accounts={','.join(user.slurm_accounts)}"
+
+    command = (
+        f"squeue {flags} --format="
+        f'"%.18i %.9P %.20j %.14u %.14a %.11T %.12M %.6D %R"'
+    )
+
+    if backend == ResourceType.SLURM_SSH:
```
zonia3000: Should I check this from `resource` or from `backend`?

fractal_server/app/routes/api/v2/job.py
```diff
@@ -228,3 +243,84 @@ async def stop_job(
...
+            except Exception as e:
+                ssh_ok = False
+                logger.error(
+                    f"Cannot establish SSH connection. Original error: {str(e)}"
+                )
+
+            if not ssh_ok:
+                raise HTTPException(
+                    status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+                    detail="Cannot establish SSH connection.",
```
zonia3000: I thought that 503 is more appropriate than 422 in this case, what do you think? Moreover, is this really needed? I copied it from similar logic but I see that `run_command` could throw `NoValidConnectionsError` too.

fractal_server/app/routes/api/v2/job.py
```diff
@@ -228,3 +243,84 @@ async def stop_job(
...
+                )
+
+            try:
+                out = fractal_ssh.run_command(cmd=command)
+                return PlainTextResponse(content=out)
+            except FractalSSHCommandError as e:
+                logger.error(f"squeue command failed. Original error: {str(e)}")
+                raise HTTPException(
+                    status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+                    detail="squeue command failed.",
```
zonia3000: I'm not sure about how to test the error condition cases. An idea would be to stop the slurmctld service to cause `squeue` to fail, but I'm not sure about the consequences. Is the same container shared across multiple tests? In that case I think it would cause issues to the other tests.

fractal_server/app/routes/api/v2/job.py
tcompa: They are meant to be identical - but this is only by construction.
Let's use the resource information.

fractal_server/app/routes/api/v2/job.py
tcompa: Agreed.

If possible, I would re-use the following function. If something else is needed, let's re-discuss it.
https://github.com/fractal-analytics-platform/fractal-server/blob/629d616c87a95ebc02d510b76b307d1e248d8a0b/fractal_server/string_tools.py#L35-L61

fractal_server/app/routes/api/v2/job.py
tcompa: As per the handling of the cannot-establish-connection branch, my first guess is that we could include it in the try/except below. We would not make the `except` specific for `FractalSSHCommandError`, but we would catch any error and then respond with an error status - without telling the user whether it's e.g. an SSH failure or a SLURM failure (which we only log, for the admin).

There's nothing wrong in separating this into SSH errors vs squeue errors, but this does not translate into something actionable for the user (apart from "wait and retry", which is valid for both kinds of errors).

fractal_server/app/routes/api/v2/job.py
tcompa: If we merge this try/except with the error-handling above (the one about SSH connection being OK), then the simple way to trigger the error would be to provide wrong SSH parameters (e.g. hostname invalid.example.org).

fractal_server/app/routes/api/v2/job.py
tcompa: An additional detail:

[`fabric.connection.Connection.run`](https://github.com/fabric/fabric/blob/ded51893f02c33d2bc7c157624c44a039a952037/fabric/connection.py#L755-L770), which is called as part of `run_command`, has a `@opens` decorator - meaning that if we skip the initial connection validation we still get an exception if the connection cannot be established.

fractal_server/app/routes/api/v2/job.py
tcompa: One more detail: `FractalSSH.run_command` already includes this step
https://github.com/fractal-analytics-platform/fractal-server/blob/629d616c87a95ebc02d510b76b307d1e248d8a0b/fractal_server/ssh/_fabric.py#L380

Is this enough?

fractal_server/app/routes/api/v2/job.py
tcompa: The typical (admittedly somewhat cumbersome) approach we follow is to define `logger = set_logger(__name__)` at the module level (note that it's `set`, rather than `get`).

fractal_server/app/routes/api/v2/job.py
zonia3000: No, as in my example you can still insert `\n` and it causes multiple commands to be executed. We can add `\n` and `\r` to the `__NOT_ALLOWED_FOR_COMMANDS__` list but maybe there are still other ways to bypass that I can't see. I was thinking about adding stricter validation in the `slurm_accounts` field. Maybe a regex like `r"^[A-Za-z0-9._-]{3,20}$"`? Which are typical `slurm_accounts` values?

tcompa: > No, as in my example you can still [..]

Got it.

> Which are typical slurm_accounts values?

Your regex is OK. If/when we hit an edge case which is not covered (possibly never..) we can refine it.

fractal_server/app/routes/api/v2/job.py
```diff
@@ -228,3 +244,60 @@ async def stop_job(
...
+                out = fractal_ssh.run_command(cmd=command)
+                return PlainTextResponse(content=out)
+        else:
+            out = execute_command_sync(command=command)
+            return PlainTextResponse(content=out)
+    except Exception as e:
+        logger.error(f"Cannot execute squeue command. Original error: {str(e)}")
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+            detail="Error executing squeue command.",
```
tcompa: Add "- please retry later."

fractal_server/app/routes/api/v2/job.py
```diff
@@ -228,3 +244,60 @@ async def stop_job(
...
+
+    flags = ""
+    if scope == "user":
+        flags = f"--user {profile.username}"
+    elif scope == "accounts" and len(user.slurm_accounts) > 0:
+        flags = f"--accounts={','.join(user.slurm_accounts)}"
+
+    command = (
+        f"squeue {flags} --format="
+        f'"%.18i %.9P %.20j %.14u %.14a %.11T %.12M %.6D %R"'
```
tcompa: New format: `squeue --format "%.12i %.9P %.24j %.14u %.14a %.11T %.12M %.6D %.4C %.10m %R"`

PR #3418, opened 2026-07-22
Explore `sqlmodel` to `sqlalchemy` migration

### LINKED ISSUE #3413: Migrate from sqlmodel to vanilla sqlalchemy
Before any further exploration, here are a few pros&cons of the migration which we may already guess:

* Pro: We make table definitions more homogeneous. Right now we have several cases where we had to use sqlalchemy directly even in a SQLModel table declaration ([example](https://github.com/search?q=repo%3Afractal-analytics-platform%2Ffractal-server%20sa_column&type=code)).
* Pro: We benefit from a large amount of on-line resources.
* Pro: In principle we could remove some tech debt by removing [a lot of vendored sqlmodel/fastapi-users code](https://github.com/fractal-analytics-platform/fractal-server/blob/main/fractal_server/app/security/__init__.py) and relying on the supported [sqlalchemy/fastapi-users support](https://fastapi-users.github.io/fastapi-users/latest/configuration/databases/sqlalchemy/).
* Con: The cost of the migration.
* Con: We may end up adding some more validation schemas (although, for the record, we typically already define and use `XXXCreate` schemas).

---

The two options we see are:
1. We rewrite all existing tables in sqlalchemy, in a way that does not require _any_ migration (or, possibly, only some trivial ones). This is the golden solution, since a new deployment would be fully transparent (restart the app in the new version, and everything works out of the box), and it does not involve any intermediate phase with coexisting solutions. Note that this likely means that we also move to the supported sqlalchemy/fastapi-users integration.
2. We proceed with incremental changes, where old tables remain in sqlmodel and new tables are in sqlalchemy only. This _may_ work, but it's a hybrid solution and it could introduce additional complexity (e.g. in testing). Existing examples:
    * https://github.com/fastapi/sqlmodel/discussions/1502
    * https://github.com/fastapi/sqlmodel/discussions/1495

Some additional references:
* Example of a large sqlalchemy project: https://github.com/polarsource/polar/tree/main/server/polar/models
* Example of a large sqlalchemy project: https://github.com/discord/access/tree/main/api/models
* Mention of a direct pydantic/sqlalchemy usage: https://pydantic.dev/docs/validation/latest/examples/orms/

---

To do: explore one of the two solutions. This is meant to be an exploration, to better assess whether a given approach is promising. The option of _not_ migrating at all should also be considered valid.

**NOTE**: If solution 2 seems to work, then we could also adopt an intermediate approach where we support both sqlalchemy and sqlamodel but we only migrate table definitions from sqlmodel to sqlalchemy one at the time - incrementally.

ychiucco:
Status update:
1. The migration in-place is mostly working, with all tests passing. Great.
2. We have not yet tried to integrate the bare sqlalchemy-fastapi-users support, which would let us drop quite a bit of legacy vendored code.
3. We need to review the model_dump re-definition. Likely it's all right, but it should be checked carefully because of how core it is.

Let's close this for now, since it is becoming stale. We can pick it up in the future, possibly in view of V3 work.

---

closes #3413 

```bash
```

Additional aspects to review:
- [x] Does a custom `model_dump` affect performance when validating a large&nested response body?
    * One option is to create a new user profile in the benchmarks, with many different task groups. Note that this should be done in `main` first.
- [x] How does typing status change? A benefit (to be verified with `ty`) seems to be that objects now have non-nullable IDs. Are there other areas where there is an improvement or a regression? Check via `ty check` on `main` and current branch.


fractal_server/app/security/__init__.py
```diff
@@ -58,118 +51,10 @@
 logger = set_logger(__name__)
 
 
-class SQLModelUserDatabaseAsync(Generic[UP, ID], BaseUserDatabase[UP, ID]):
```
tcompa, line 61: Let's compare the original version with this one, before fully replacing it. If they are identical, we can more confidently move to `SQLAlchemyUserDatabase`

fractal_server/app/models/base.py
```diff
@@ -0,0 +1,46 @@
...
+            set_names -= exclude
+        return {name: self.__dict__[name] for name in set_names}
+
+    def model_dump_json(
+        self,
+        *,
+        include: set[str] | None = None,
+        exclude: set[str] | None = None,
+    ) -> str:
+        return pydantic_core.to_json(
```
tcompa: Let's replace with `json.dumps`.

Why `decode`?

fractal_server/app/models/base.py
```diff
@@ -0,0 +1,46 @@
...
+        A lot of code relies on calling `.model_dump()`/`.model_dump_json()`
+        directly on ORM instances, mirroring their former pydantic-based
+        behavior: only mapped columns are included (never `relationship()`
+        attributes), and only columns that already have a concrete value in
+        `__dict__` (i.e. loaded from the DB, or explicitly assigned) are
+        included, omitting columns whose value is still pending a
+        server-side default on a not-yet-flushed instance.
+        """
+        column_names = {c.key for c in self.__mapper__.column_attrs}
+        set_names = column_names & self.__dict__.keys()
```
tcompa: Does `dict(self)` work?

fractal_server/app/models/base.py
```diff
@@ -0,0 +1,46 @@
...
+
+        A lot of code relies on calling `.model_dump()`/`.model_dump_json()`
+        directly on ORM instances, mirroring their former pydantic-based
+        behavior: only mapped columns are included (never `relationship()`
+        attributes), and only columns that already have a concrete value in
+        `__dict__` (i.e. loaded from the DB, or explicitly assigned) are
+        included, omitting columns whose value is still pending a
+        server-side default on a not-yet-flushed instance.
+        """
+        column_names = {c.key for c in self.__mapper__.column_attrs}
```
tcompa: Could we do this without private attributes?

fractal_server/app/models/base.py
ychiucco: I've replaced the call to `pydantic_core` with the following:
```python
def model_dump_json(
    self,
    *,
    include: set[str] | None = None,
    exclude: set[str] | None = None,
) -> str:
    return (
        TypeAdapter(dict[str, Any])
        .dump_json(self.model_dump(include=include, exclude=exclude))
        .decode()
    )
```

---

The problem with `json.dumps` is the serialization of non-built-in types:
```python
import json
from datetime import datetime, timezone

json.dumps({"a": 1, "timestamp_created": datetime.now(timezone.utc)})
```
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.12/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/encoder.py", line 200, in encode
    chunks = self.iterencode(o, _one_shot=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/encoder.py", line 258, in iterencode
    return _iterencode(o, 0)
           ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/json/encoder.py", line 180, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type datetime is not JSON serializable
```

With `pydantic.TypeAdapter` instead:

```python
from pydantic import TypeAdapter
from typing import Any
from datetime import datetime, timezone

TypeAdapter(dict[str, Any]).dump_json({"a": 1, "timestamp_created": datetime.now(timezone.utc)})
```
```
b'{"a":1,"timestamp_created":"2026-09-01T13:56:46.400406Z"}'
```

---

`decode` is still needed because `TypeAdapter.dump_json` returns `bytes`.

fractal_server/app/models/base.py
ychiucco: `dict(self)` isn't possible because ORM model instances aren't iterable or mapping-like (no `__iter__`, no `keys()/__getitem__`), so `dict()` has nothing to convert them with:
```
TypeError: 'ProjectV2' object is not iterable
```

Even if we bypassed that by iterating `self.__dict__` instead, that dict mixes in SQLAlchemy's internal `_sa_instance_state` (and any loaded relationships), so it isn't "just the columns" either.

tcompa:
RE:
> [ ] Does a custom `model_dump` affect performance when validating a large&nested response body?

note that

> When creating an instance of [TypeAdapter](https://pydantic.dev/docs/validation/latest/api/pydantic/type_adapter/#pydantic.type_adapter.TypeAdapter), the provided type must be analyzed and converted into a pydantic-core schema. This comes with some non-trivial overhead, so it is recommended to create a TypeAdapter for a given type just once and reuse it in loops or other performance-critical code.
> (https://pydantic.dev/docs/validation/latest/concepts/type_adapter/)

fractal_server/app/models/base.py
```diff
@@ -0,0 +1,49 @@
...
+from sqlalchemy import inspect
+from sqlalchemy.orm import DeclarativeBase
+
+from fractal_server.migrations.naming_convention import NAMING_CONVENTION
+
+
+class Base(DeclarativeBase):
+    metadata = MetaData(naming_convention=NAMING_CONVENTION)
+
+    def model_dump(
```
tcompa: Independently on the open questions about performance comparison between this method and `main`, I suggest we rename `model_dump` and `model_dump_json`, and that we move them from `Base` methods into standalone functions. The rationale is that calling the `model_dump` method of an ORM model is really a `SQLModel` idiom, and it would be mixed in the codebase with usage of the actual original `model_dump` method (for Pydantic objects).

fractal_server/app/models/base.py
```diff
@@ -0,0 +1,49 @@
...
+        A lot of code relies on calling `.model_dump()`/`.model_dump_json()`
+        directly on ORM instances, mirroring their former pydantic-based
+        behavior: only mapped columns are included (never `relationship()`
+        attributes), and only columns that already have a concrete value in
+        `__dict__` (i.e. loaded from the DB, or explicitly assigned) are
+        included, omitting columns whose value is still pending a
+        server-side default on a not-yet-flushed instance.
+        """
+        column_names = {c.key for c in inspect(self).mapper.column_attrs}
+        set_names = column_names & self.__dict__.keys()
```
tcompa: This feature (only including keys which are one of the column names and which are part of `__dict__`) must be tested carefully, e.g. because in the future we may start using aliases and it's not clear whether it would be supported (to be clear: there is no need for supporting column-name aliases right now).

fractal_server/app/models/base.py
```diff
@@ -0,0 +1,49 @@
...
+        return {name: self.__dict__[name] for name in set_names}
+
+    def model_dump_json(
+        self,
+        *,
+        include: set[str] | None = None,
+        exclude: set[str] | None = None,
+    ) -> str:
+        return (
+            TypeAdapter(dict[str, Any])
```
tcompa: This feature (using `TypeAdapter` with a generic dictionary) needs unit tests for a few typical behaviors (datetime.datetime, pathlib.Path, uuid.UUID, ...), including cases of nested objects - where I am not 100% sure of what behavior we should support. The existing examples of non-scalar database columns are JSON and arrays, as far as I can tell. How would e.g. an array of timestamps be serialized here?

fractal_server/app/models/base.py
tcompa: Re-opening, to stress

> we move them from `Base` methods into standalone functions.

fractal_server/app/routes/api/v2/submit.py
```diff
@@ -214,27 +215,29 @@ async def submit_job(
...
         dataset_dump=json.loads(
-            dataset.dump_model_to_json(
-                exclude={"images", "history", "is_starred"}
+            dump_model_to_json(
+                dataset, exclude={"images", "history", "is_starred"}
             )
         ),
         workflow_dump=json.loads(
-            workflow.dump_model_to_json(
+            dump_model_to_json(
```
tcompa, line 223: Is the round-trip required?

tests/no_version/test_unit_models_base.py
```diff
@@ -0,0 +1,132 @@
...
+    dumped = hr.dump_model()
+    assert dumped["status"] == "done"
+    assert dumped["num_available_images"] == 5
+    assert dumped["timestamp_started"] == datetime(2023, 1, 1, 12, 0, 0)
+
+
+def test_dump_model_to_json_datetime_is_isoformat():
+    hr = _history_run()
+    dumped = json.loads(hr.dump_model_to_json())
+    assert dumped["timestamp_started"] == "2023-01-01T12:00:00"
```
tcompa, line 48: On the right-hand side, write a datetime with the proper serialization method - so that we are aware of which one is being used.

tests/no_version/test_unit_models_base.py
```diff
@@ -0,0 +1,132 @@
...
+    `datetime` column is: as an ISO-8601 string.
+    """
+    hr = _history_run(
+        workflowtask_dump={
+            "timestamps": [
+                datetime(2023, 1, 1),
+                datetime(2023, 1, 2, 8, 30),
+            ]
+        },
+    )
```
tcompa, line 91: Question:
If we commit `hr` at this point and then refresh it and read it - what is the value of `timestamps`? My guess is that it is already serialized.
If so, the proper example requires a new table with an array of timestamps.

tests/no_version/test_unit_models_base.py
```diff
@@ -0,0 +1,132 @@
...
+def test_dump_model_to_json_uuid_and_path_in_json_column():
+    """
+    `UUID`/`Path` values are not used by any current column, but if one
+    ever ends up inside a `dict[str, Any]`/JSON column, it is serialized
+    to its string representation - same as pydantic's `model_dump_json`.
+    """
+    some_uuid = uuid.uuid4()
+    hr = _history_run(
+        workflowtask_dump={"id": some_uuid, "path": Path("/tmp/foo/bar")},
+    )
```
tcompa, line 108: Same as above: who is actually serializing the UUID?

fractal_server/app/models/base.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def dump_model_to_json(
+    obj: Base,
+    *,
+    include: set[str] | None = None,
+    exclude: set[str] | None = None,
+) -> str:
+    return (
+        TypeAdapter(dict[str, Any])
```
tcompa, line 48: When creating an instance of [TypeAdapter](https://pydantic.dev/docs/validation/latest/api/pydantic/type_adapter/#pydantic.type_adapter.TypeAdapter), the provided type must be analyzed and converted into a pydantic-core schema. This comes with some non-trivial overhead, so it is recommended to create a TypeAdapter for a given type just once and reuse it in loops or other performance-critical code.
    (https://pydantic.dev/docs/validation/latest/concepts/type_adapter/)

ychiucco:
> How does typing status change? A benefit (to be verified with ty) seems to be that objects now have non-nullable IDs. Are there other areas where there is an improvement or a regression? Check via ty check on main and current branch.

```
$ git switch 3413-explore-sqlmodel-sqlalchemy-migration

$ uvx ty check | tail -1
Found 1132 diagnostics

$ uvx ty check --output-format concise 2>&1 | grep -oP 'error\[\K[^\]]+' | sort | uniq -c | sort -rn
    877 invalid-argument-type
     74 unresolved-attribute
     60 invalid-assignment
     40 unresolved-import
     26 invalid-return-type
     14 missing-argument
     11 unsupported-operator
      8 not-subscriptable
      7 unknown-argument
      5 invalid-type-arguments
      4 no-matching-overload
      2 call-non-callable
      1 not-iterable
      1 invalid-parameter-default
      1 invalid-method-override
```

```
$ git switch main

$ uvx ty check | tail -1
Found 2463 diagnostics

$ uvx ty check --output-format concise 2>&1 | grep -oP 'error\[\K[^\]]+' | sort | uniq -c | sort -rn
   1837 invalid-argument-type
    217 missing-argument
    191 unresolved-attribute
     73 invalid-assignment
     40 unresolved-import
     32 no-matching-overload
     28 invalid-return-type
     22 unsupported-operator
      8 not-subscriptable
      7 unknown-argument
      3 invalid-method-override
      2 call-non-callable
      1 not-iterable
      1 invalid-parameter-default
```

| error | main | 3413 | Δ |
|---|---:|---:|---:|
| invalid-argument-type | 1837 | 877 | -960 |
| missing-argument | 217 | 14 | -203 |
| unresolved-attribute | 191 | 74 | -117 |
| invalid-assignment | 73 | 60 | -13 |
| unresolved-import | 40 | 40 | 0 |
| no-matching-overload | 32 | 4 | -28 |
| invalid-return-type | 28 | 26 | -2 |
| unsupported-operator | 22 | 11 | -11 |
| not-subscriptable | 8 | 8 | 0 |
| unknown-argument | 7 | 7 | 0 |
| invalid-method-override | 3 | 1 | -2 |
| call-non-callable | 2 | 2 | 0 |
| not-iterable | 1 | 1 | 0 |
| invalid-parameter-default | 1 | 1 | 0 |
| invalid-type-arguments | 0 | 5 | +5 |
| **Total** | **2463** | **1132** | **-1331** |

tcompa:
> > How does typing status change? A benefit (to be verified with ty) seems to be that objects now have non-nullable IDs. Are there other areas where there is an improvement or a regression? Check via ty check on main and current branch.
> 
> ```
> $ git switch 3413-explore-sqlmodel-sqlalchemy-migration
> 
> $ uvx ty check | tail -1
> Found 1132 diagnostics
> 
> $ uvx ty check --output-format concise 2>&1 | grep -oP 'error\[\K[^\]]+' | sort | uniq -c | sort -rn
>     877 invalid-argument-type
>      74 unresolved-attribute
>      60 invalid-assignment
>      40 unresolved-import
>      26 invalid-return-type
>      14 missing-argument
>      11 unsupported-operator
>       8 not-subscriptable
>       7 unknown-argument
>       5 invalid-type-arguments
>       4 no-matching-overload
>       2 call-non-callable
>       1 not-iterable
>       1 invalid-parameter-default
>       1 invalid-method-override
> ```
> 
> ```
> $ git switch main
> 
> $ uvx ty check | tail -1
> Found 2463 diagnostics
> 
> $ uvx ty check --output-format concise 2>&1 | grep -oP 'error\[\K[^\]]+' | sort | uniq -c | sort -rn
>    1837 invalid-argument-type
>     217 missing-argument
>     191 unresolved-attribute
>      73 invalid-assignment
>      40 unresolved-import
>      32 no-matching-overload
>      28 invalid-return-type
>      22 unsupported-operator
>       8 not-subscriptable
>       7 unknown-argument
>       3 invalid-method-override
>       2 call-non-callable
>       1 not-iterable
>       1 invalid-parameter-default
> ```
> 
> error 	main 	3413 	Δ
> invalid-argument-type 	1837 	877 	-960
> missing-argument 	217 	14 	-203
> unresolved-attribute 	191 	74 	-117
> invalid-assignment 	73 	60 	-13
> unresolved-import 	40 	40 	0
> no-matching-overload 	32 	4 	-28
> invalid-return-type 	28 	26 	-2
> unsupported-operator 	22 	11 	-11
> not-subscriptable 	8 	8 	0
> unknown-argument 	7 	7 	0
> invalid-method-override 	3 	1 	-2
> call-non-callable 	2 	2 	0
> not-iterable 	1 	1 	0
> invalid-parameter-default 	1 	1 	0
> invalid-type-arguments 	0 	5 	+5
> **Total** 	**2463** 	**1132** 	**-1331**

Looks great!

fractal_server/app/db/__init__.py
```diff
@@ -22,6 +19,20 @@
...
+
+def _json_serializer(obj: Any) -> str:
+    """
+    Serializer for `JSON`/`JSONB` columns, used in place of plain
+    `json.dumps` so that values which are not natively JSON-serializable
+    (e.g. `datetime`, `UUID`, `Path`) but can appear nested in a
+    `dict[str, Any]` column are encoded the same way as in
+    `dump_model_to_json` (pydantic's `model_dump_json` semantics).
+    """
+    return AnyTypeAdapter.dump_json(obj).decode()
```
tcompa: Do we have a list of all types that would require this custom serialization? Is it only for `datetime`s and `Path`s? Perhaps UUIDs? Anything else?





If it is only for datetimes, then I think an alternative (pydantic-independent) solution would be
```python
from json import JSONEncoder

class CustomJSONEncoder(JSONEncoder):
    def default(self, obj):
        if isinstance(obj, datetime):
            return obj.isoformat()
        elif isinstance(obj, Path):
            return obj.as_posix()
        else:
            return super().default(obj)
```

Questions:
1. Would it work?
2. How does it compare to the `TypeAdapter` one in terms of performance for some heavy/nested object? This can be tested in a very isolated benchmark, since the two options would share an identical API.
3. How does the best option of the two compare to the `sqlmodel`-based approach?

fractal_server/app/db/__init__.py
```diff
@@ -22,6 +19,20 @@
 logger = set_logger(__name__)
 
 
+AnyTypeAdapter = TypeAdapter(Any)
```
tcompa: I am confused by why we need this adapter/serializer, even though tests were already passing before adding it. Let's rediscuss this (live, likely).

tcompa:
Recap of useful benchmarks

1. Isolated test of JSON serializers: customer JSONEncoder vs TypeAdapter-based serialization. Test case: a large nested dictionary with datetimes, `Path`s, `UUID`s and some JSON-compatible types (dict, list, string, null). Order of magnitude: a total of 10_000 nodes in the graph. If JSONEncoder is at least comparable, that's the preference we have. This test does not import sqlalchemy.
2. Is `dump_model_to_json` needed at all? Note that it always called in a round trip as in `json.loads(dump_model_to_json(..))`. Can we use `dump_model` directly?
3. A new test needs to run both on this branch and on main.
    * On main it calls `model_dump` on a `SQLModel` object (as larger as possible, and with some include/exclude). Perhaps the best is to use a dedicated new table, with 20-30 columns and some of them being some large JSON.
    * On the current branch, we re-create the same table and run the same script, but with `dump_model` called on a `DeclarativeBase` object.

Expectations:
1. Test 1 shows a significant performance difference.
2. `dump_model_to_json` can be removed
3. Test 3 shows a minimal performance difference.

tcompa:
After https://github.com/fractal-analytics-platform/fractal-server/pull/3418#issuecomment-5712277962 is complete:
- [ ] Review all model definitions once more
- [ ] Review the integration with fastapi-users once more

And basically that's it.

ychiucco:
>     1. Isolated test of JSON serializers: customer JSONEncoder vs TypeAdapter-based serialization. Test case: a large nested dictionary with datetimes, `Path`s, `UUID`s and some JSON-compatible types (dict, list, string, null). Order of magnitude: a total of 10_000 nodes in the graph. If JSONEncoder is at least comparable, that's the preference we have. This test does not import sqlalchemy.

After https://github.com/fractal-analytics-platform/fractal-server/pull/3418/commits/17d119699c6f0ea54510c7547102ca34ed6c3c52
```
uv run python benchmarks/bench_json_serializer.py
Generated nested structure with 10000 nodes.

Timing each serializer over 20 repetitions:

 json.dumps(default=...): mean=   1.814 ms  median=   1.789 ms  min=   1.738 ms  max=   2.115 ms
        TypeAdapter(Any): mean=   0.984 ms  median=   0.970 ms  min=   0.955 ms  max=   1.201 ms

json.dumps / TypeAdapter time ratio: 1.84x
```

tcompa:
> > ```
> > 1. Isolated test of JSON serializers: customer JSONEncoder vs TypeAdapter-based serialization. Test case: a large nested dictionary with datetimes, `Path`s, `UUID`s and some JSON-compatible types (dict, list, string, null). Order of magnitude: a total of 10_000 nodes in the graph. If JSONEncoder is at least comparable, that's the preference we have. This test does not import sqlalchemy.
> > ```
> 
> After [17d1196](https://github.com/fractal-analytics-platform/fractal-server/commit/17d119699c6f0ea54510c7547102ca34ed6c3c52)
> 
> ```
> uv run python benchmarks/bench_json_serializer.py
> Generated nested structure with 10000 nodes.
> 
> Timing each serializer over 20 repetitions:
> 
>  json.dumps(default=...): mean=   1.814 ms  median=   1.789 ms  min=   1.738 ms  max=   2.115 ms
>         TypeAdapter(Any): mean=   0.984 ms  median=   0.970 ms  min=   0.955 ms  max=   1.201 ms
> 
> json.dumps / TypeAdapter time ratio: 1.84x
> ```

Great, thanks.

Here is a quick scaling analysis:

```console
$ uv run python benchmarks/bench_json_serializer.py 
Generated nested structure with 100 nodes.

Timing each serializer over 20 repetitions:

 json.dumps(default=...): mean=   0.036 ms  median=   0.035 ms  min=   0.030 ms  max=   0.053 ms
        TypeAdapter(Any): mean=   0.014 ms  median=   0.013 ms  min=   0.013 ms  max=   0.020 ms

json.dumps / TypeAdapter time ratio: 2.58x

$ uv run python benchmarks/bench_json_serializer.py 
Generated nested structure with 1000 nodes.

Timing each serializer over 20 repetitions:

 json.dumps(default=...): mean=   0.630 ms  median=   0.630 ms  min=   0.592 ms  max=   0.682 ms
        TypeAdapter(Any): mean=   0.150 ms  median=   0.145 ms  min=   0.132 ms  max=   0.268 ms

json.dumps / TypeAdapter time ratio: 4.19x

$ uv run python benchmarks/bench_json_serializer.py 
Generated nested structure with 10000 nodes.

Timing each serializer over 20 repetitions:

 json.dumps(default=...): mean=   3.236 ms  median=   3.220 ms  min=   3.060 ms  max=   3.446 ms
        TypeAdapter(Any): mean=   1.451 ms  median=   1.421 ms  min=   1.346 ms  max=   1.671 ms

json.dumps / TypeAdapter time ratio: 2.23x

$ uv run python benchmarks/bench_json_serializer.py 
Generated nested structure with 100000 nodes.

Timing each serializer over 20 repetitions:

 json.dumps(default=...): mean=  32.162 ms  median=  32.113 ms  min=  31.782 ms  max=  32.867 ms
        TypeAdapter(Any): mean=  14.613 ms  median=  14.557 ms  min=  14.244 ms  max=  15.201 ms

json.dumps / TypeAdapter time ratio: 2.20x
```

Up to 10_000 nodes we are in the millisecond range, which is not relevant on any API request handling nor on any in-runner operation. When we reach 100_000 nodes (which already seems quite large) we start to see tens of milliseconds, but the slowdown factor remains around 2x.

--> Let's keep the standard-library option, and add a comment about the possible optimization through `TypeAdapter `.

tcompa:
Side comment: I'd suggest to use a custom encoder rather than a custom default method, as in 6412657894032004ddd39aab6bee8420a41df4b6 - but this is just a preference and I don't have any strong motivation about it.

tcompa:
Copying from https://raw.githubusercontent.com/fractal-analytics-platform/fractal-server/refs/heads/3413-explore-sqlmodel-sqlalchemy-migration/benchmarks/json_serialization/README.md:




## (1) Object serialization to JSON

A customized `json.dumps` function (which handles some special binary types) is used twice in the PR #318:
* As the `json_serializer` argument for SQLAlchemy-engine creation, so that JSON/JSONB colums can be created in the ORM by providing Python objects (including with attributes of type `datetime.datetime`, `uuid.UUID` or `pathlib.Path`).
* As part of the function that replaces `SQLModel.model_dump` (see section 2 below)

The test can be run as in
```bash
uv run bench_json_serializer.py
```

An example of output (as of cef0a63af4) looks like
```
Generated nested structure with 100 nodes.
[ json.dumps]: mean=0.07 ms  median=0.07 ms  min=0.07 ms  max=0.10 ms
[TypeAdapter]: mean=0.03 ms  median=0.03 ms  min=0.03 ms  max=0.04 ms
TypeAdapter speed-up ratio: 2.65x

Generated nested structure with 1000 nodes.
[ json.dumps]: mean=0.49 ms  median=0.56 ms  min=0.29 ms  max=0.60 ms
[TypeAdapter]: mean=0.18 ms  median=0.15 ms  min=0.13 ms  max=0.27 ms
TypeAdapter speed-up ratio: 2.67x

Generated nested structure with 10102 nodes.
[ json.dumps]: mean=2.90 ms  median=2.86 ms  min=2.84 ms  max=3.18 ms
[TypeAdapter]: mean=1.41 ms  median=1.40 ms  min=1.39 ms  max=1.46 ms
TypeAdapter speed-up ratio: 2.07x

Generated nested structure with 100000 nodes.
[ json.dumps]: mean=28.97 ms  median=28.89 ms  min=28.60 ms  max=30.44 ms
[TypeAdapter]: mean=13.48 ms  median=13.18 ms  min=13.07 ms  max=21.23 ms
TypeAdapter speed-up ratio: 2.15x
```

tcompa:
> A new test needs to run both on this branch and on main.
> 
>     * On main it calls `model_dump` on a `SQLModel` object (as larger as possible, and with some include/exclude). Perhaps the best is to use a dedicated new table, with 20-30 columns and some of them being some large JSON.
> 
>     * On the current branch, we re-create the same table and run the same script, but with `dump_model` called on a `DeclarativeBase` object.

---

Copying from https://raw.githubusercontent.com/fractal-analytics-platform/fractal-server/refs/heads/3413-explore-sqlmodel-sqlalchemy-migration/benchmarks/json_serialization/README.md:

ORM objects of the `sqlmodel.SQLModel` type have a `model_dump` method based on `pydantic`, which is sometimes used in `fractal-server` to produce a Python-dictionary version of the ORM object.
As part of https://github.com/fractal-analytics-platform/fractal-server/pull/3418, we implemented an equivalent function for `sqlalchemy.DeclarativeBase` objects.

> **Note**: In both cases, the resulting dictionary may not be JSON-serializable by a vanilla `json.dumps`, e.g. because it contains some `pathlib.Path` or `datetime.datetime` objects.

Here we compare the former method (by working on the current `main` branch) and the new one (by working in the current branch as part of https://github.com/fractal-analytics-platform/fractal-server/pull/3418).

Note that we use the database-committed objects. This is to avoid a discrepancy about when defaults are populated (which for SQLModel takes place uponc object construction, while for SQLAlchemy takes place in the databse).

```
$ pwd
/redacted/fractal-server/benchmarks/json_serialization

$ git show --oneline -s
5a5c1f3418 (HEAD -> 3413-explore-sqlmodel-sqlalchemy-migration) Use proper SQLModel base

$ dropdb --if-exists json-benchmarks

$ createdb json-benchmarks

$ POSTGRES_HOST=/var/run/postgresql POSTGRES_DB=json-benchmarks JWT_SECRET_KEY=123 uv run python bench_orm_object_dump_sqlalchemy.py
      Built fractal-server @ file:///home/tommaso/Fractal/fractal-server                                                                                                                                                         Uninstalled 1 package in 0.35ms
Installed 1 package in 1ms
[   Resource]: mean=2.38 ns  median=2.33 ns  min=2.24 ns  max=3.01 ns
[    Profile]: mean=2.15 ns  median=1.96 ns  min=1.90 ns  max=5.03 ns
[  UserOAuth]: mean=2.32 ns  median=2.18 ns  min=1.94 ns  max=4.45 ns
[TaskGroupV2]: mean=3.67 ns  median=3.24 ns  min=3.02 ns  max=7.45 ns

$ git checkout main ../../fractal_server/
Updated 85 paths from 80623de687

$ POSTGRES_HOST=/var/run/postgresql POSTGRES_DB=json-benchmarks JWT_SECRET_KEY=123 uv run python bench_orm_object_dump_sqlmodel.py
[   Resource]: mean=2.80 ns  median=2.70 ns  min=2.64 ns  max=3.67 ns
[    Profile]: mean=2.28 ns  median=2.11 ns  min=2.03 ns  max=5.20 ns
[  UserOAuth]: mean=2.53 ns  median=2.38 ns  min=2.26 ns  max=5.09 ns
[TaskGroupV2]: mean=6.67 ns  median=6.43 ns  min=6.30 ns  max=10.56 ns


$ git restore --staged ../../fractal_server/

$ git restore  ../../fractal_server/

$ git status
On branch 3413-explore-sqlmodel-sqlalchemy-migration
Your branch is ahead of 'origin/3413-explore-sqlmodel-sqlalchemy-migration' by 17 commits.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean
```

---

Times are in the nanoseconds range, meaning this benchmark won't make any real-life difference. Nevertheless, we observe that the new version is even faster than the older one.

tcompa:
Closing in favor of #3438, #3439 and #3440 - to make reviews simpler.

PR #3428, opened 2026-08-25
Add pylock file and update pre-commit hooks

### LINKED ISSUE #3427: Adopt dependency locks
The main issue we want to solve here is how to lock `fractal-server` dependencies for a production instance.

The benefits are quite obvious, but just to be clear:
* It would make different instances more homogeneous (e.g. `fractal-server` version X.Y.Z would automatically imply sqlalchemy version A.B.C).
* It would reduce dependency-related issues (which are already already quite rare - due to the fact that we have _some_ upper bounds for all direct dependencies).
* It would provide immediately-readable information in case of vulnerabilities in some dependencies, including transitive ones.
* It would make it simple to provide and check hashes.
* It may make it simple to provide and check digital attestations (see https://blog.pypi.org/posts/2024-11-14-pypi-now-supports-digital-attestations or https://blog.trailofbits.com/2024/11/14/attestations-a-new-generation-of-signatures-on-pypi/, and https://snarky.ca/why-pylock-toml-includes-digital-attestations/ for how this may matter).
* It would make the usage of `uv.lock` much more meaningful, by reproducing the same environment for a production instance.

## Two options

While maintaining the current assumption that a `fractal-server` deployment must be possible via `pip install` in a fresh virtual environment (that is, not necessarily through a specific Python tool and/or containerized builds), we are left with two main options:
* [Standard Python lockfiles](https://packaging.python.org/en/latest/specifications/pylock-toml/#pylock-toml-spec), AKA pylock
* `--constraint` specifications for `pip install`, as [done e.g. by Apache Airflow](https://airflow.apache.org/docs/apache-airflow/stable/installation/installing-from-pypi.html#reproducible-airflow-installation).

The `--constraint` option is simpler and based on a more mature `pip` feature, but for the moment we will only consider the pylock one. Reasons:
* This is also a narrow-scope version of a potential future discussion about using locks for task collection (rather than for Fractal services). It'd be valuable to gain some expertise before tackling the broader discussion (task packages may come with extras and/or with additional system-dependent requirements, and in general they will have more dependencies - both in number and diversity).
* Tooling may evolve in directions that are relevant for the Fractal-service use case, e.g. re: attestations (see https://github.com/trailofbits/are-we-pep740-yet/issues/35, https://github.com/trailofbits/pylock-attestations and https://github.com/astral-sh/uv/issues/9122). These features would be natively present in pylock, while I think the `--constraint` option is meant to be a simpler one.
* As far as I understand `--constraint` is a `pip` feature rather than a standard one.

## pylock

As of [pip 26.1](https://pip.pypa.io/en/stable/news/#v26-1) (released in April 2026), there is experimental support for `pip install -r /some/pylock.toml`. We want to find a way to maintain/distribute/consume an up-to-date lock file to enable locked-dependencies installs.

Here is a plan on how to proceed.

### How to produce the lock file

* Since we already use `uv`, we should extract the pylock file from the `uv.lock` file, as in `uv export --format pylock.toml`.
* We should not include development/documentation dependencies (that is, we should use `--no-group dev --no-group docs`)
* We should not include the `fractal-server` package itself, since it would be marked as an editable package (given that the pylock would be produced before publishing on pypi, and pypi must remain the single authoritative source for the fractal-server package). This means adding the `--no-emit-project` option.

The final command should look something like
```
uv export --format pylock.toml --no-group dev --no-group docs --no-emit-project -o pylock.prod.toml
```

### How to distribute the lock file

Ideally, there would exist a standard way to distribute a lock file for a Python application - e.g. on PyPI. This is not the case yet, see e.g. the long discussions in https://discuss.python.org/t/pre-pep-include-pylock-toml-files-inside-wheels/99385/ or https://discuss.python.org/t/pre-pep-add-ability-to-install-a-package-with-reproducible-dependencies/99497.

The current best solution we can come up with would then be to commit the file to the `git` repository, and add automation to make sure that it is up to date. Note that the goal is to have the correct lock file _at every commit_, rather than e.g. updating it after a release. That is, we would like the committed pylock file for tag A.B.C to be the one corresponding to the dependencies tested in the same commit that then published version A.B.C to PyPI.

We can identify the best approach during implementation, with obvious options being a `pre-commit` rule and/or a dedicated GitHub action.

> Side note: the apache/airflow approach of using multiple orphan branches to store constraint files corresponding to different versions seems too complex on the `git` side, and we would rather stick with "every branch/tag always has an up-to-date lockfile committed").

### How to consume the lock file

The installation script would look like
```bash
# Specify target version
FRACTAL_SERVER_VERSION="2.24.2"

# Create a fresh environment
rm -rf venv
python3 -m venv venv
source venv/bin/activate

# Update pip (which is needed to support installing from a lock file)
pip install pip -U

# Install dependencies as specified in pylock
rm -f pylock.prod.toml
wget https://raw.githubusercontent.com/fractal-analytics-platform/fractal-server/refs/tags/${FRACTAL_SERVER_VERSION}/pylock.prod.toml
pip install -r ./pylock.prod.toml

# Install fractal-server
pip install "fractal-server==${FRACTAL_SERVER_VERSION}" --no-deps
```

Note that this feature remains fully optional. By commenting out the `pip install -r ./pylock.prod.toml` step, one would get the set of (non-locked) dependencies that the `pip` resolver would identify at the specific time of the `pip install fractal-server` command.

Also note that the final `--no-deps` is not required (because dependencies are already installed, and they would not be re-installed twice), but it makes the script a bit more readable. One may also drop it, in favor of double-checking that all dependencies are there.

tcompa:

- [x] I reviewed the question about Python versions in the lockfile.
```bash
```


tcompa:
Converting to draft due to the following question:
Does the Python version used for generating the lock matter? I guess so.
Do we have to generate locks for each supported Python version? I also guess so.

ychiucco (approved):

